PR #8, opened 2012-07-17
Easier API for Shape creation

padolsey:
**(please discuss)**

Creating shapes is something that users will want to do very often. Filling a shape and applying a stroke are also pretty frequent use-cases. 

Currently `Shape` instances only support the generalised `attr` method with no convenience methods for the most common  use-cases.

This pull request would make shape creation look like this:

``` js
var a = rect(0, 0, 50, 50).fill('red').line('green', 2);
```

Instead of the current way:

``` js
var a = Shape.rect(0, 0, 50, 50).attr({
  fillColor: 'red',
  lineColor: 'green',
  lineWidth: 2
});
```

Three main changes:
- All `Shape.*` factories are exposed globally, including `rect`, `arc`, `star`, `circle` etc. (this might create annoying global conflicts for the user)
- `Shape#fill()`: can accept a color, a gradient or a bitmap as the fill.
- `Shape#line()`: can accept a color or a gradient, and a lineWidth as the 2nd argument.

Both `line()` and `fill()` are just setters. They're not getters. It would be impossible to implement them as getters because a shape can have multiple fill styles (a fillColor and a fillGradient, for example).

Exposing all factories might be a little bit too much though, as it further pollutes the global namespace. I'm okay to remove that... just wanted to see what people thought.

---

My reasoning is that we should provide convenient ways of doing the most common things. Currently the shape creation process is needlessly verbose.

### DIFF
--- a/src/runner/environment.js
+++ b/src/runner/environment.js
@@ -66,7 +66,13 @@ define([
       easing: easing,
       filter: filter,
       stage: stage,
-      version: version
+      version: version,
+      arc: Shape.arc,
+      circle: Shape.circle,
+      ellipse: Shape.ellipse,
+      rect: Shape.rect,
+      polygon: Shape.polygon,
+      star: Shape.star
     };
 
     this.assetLoader = assetLoader;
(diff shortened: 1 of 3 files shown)


davidaurelio:
Why not? Not sure if we are introducing to many global symbols, but I like the idea

PR #9, opened 2012-07-24
Path special-classes (plus renaming `Shape` to `Path`)

padolsey:
- Replaces the `Shape` class with the `Path` class.
- Introduces new classes for all shape types, namely:
  - Polygon
  - Ellipse
  - Circle
  - Rect
  - Star
  - Arc

`Path.*` factories still exist, but don't return their special path class instances. In other words, `Path.rect()` will just return a `Path` instance (like before), not a `Rect` instance. This is to retain previous functionality, most significantly shape-morphing ([see this](https://github.com/uxebu/bonsai/commit/4877991e61051da039daf3ad757ecda66ae1eeb0)).

All special path classes (Arc, Rect etc.) inherit from `SpecialPath`. SpecialPath directly subclasses Path and provides a convenience argument for adding special attributes for unique shape types. See the implementations of `Rect`, `Arc` (etc.) for demonstrations of this convenience argument. In addition `SpecialPath` removes the `morphTo` method ([see this](https://github.com/uxebu/bonsai/commit/4877991e61051da039daf3ad757ecda66ae1eeb0)) and makes sure the path is redrawn (using the `_make` method) when attributes are changed (e.g. when `radius` is changed on a `Circle` instance).

Individual path classes each have their applicable properties as regular bonsai attributes. So you can do stuff like:

``` js
new Rect(0,0,100,100).addTo(stage).animate('1s', {
  width: 200,
  height: 200,
  cornerRadius: 20
});
new Star(100, 100, 50, 5, 2)
  .attr({ rays: 16 })
  .fill('blue')
  .addTo(stage)
  .animate('1s', {
    factor: 1
  });
```

Overview of new special attributes:
- `Rect`: width, height, cornerRadius
- `Star`: radius, rays, factor
- `Circle`: radius
- `Ellipse`: radiusX, radiusY
- `Polygon`: radius, sides
- `Arc`: radius, startAngle, endAngle, antiClockwise


basecode:
For what I see here is that we need to maintain two different APIs for "convenient path wrappers", or more specific `Path.rect` vs `Rect`. Do you think this is necessary? I think this is confusing and triggers questions about what wrapper is better. I see the "morph" problem but maybe we can get around it. I would suggest to remove the "old" wrappers and go with the SpecialPaths. I think we could make it possible to morph SpecialPaths by introducing new API.

`morphTo` wouldn't work because we can't return a "modified SpecialPath" thing. But what about a `morph` function that returns a new instance of `Path`?

`var myPath = morph(new Rect(…)).to(new Circle(…)).animate(…)`

1) `morph` just clones a SpecialPath and converts it to a Path.
2) `to` overwrites the path-segments of path (that was created in step 1)
3) `animate` triggers an animation that changes one path into another through a seamless transition

padolsey:
Thanks for your comment Tobi. I agree with what you've said.

I chose not to do jsdoc blocks for the factories (`Path.*`) because I think they should be deprecated ASAP. Currently we do have some things that rely on the old behaviour (morph examples), so I'd like to keep the factories for now, at least until we make a new API for morphing SpecialPath classes.

I don't think creating that new morphing API should happen in this branch though. Plus I'm not sure how much effort it'll take.

padolsey:
@basecode Do you think it's okay to merge?

FWIW, currently the factories are only being used in a couple of places for the morphing examples. I hope we can look at making another morphing API for SpecialPath once this pull request is merged.

PR #25, opened 2012-08-20
Fix pattern image fill for Firefox

klipstein:
@basecode: using images as pattern fill is currently leading to strange artifacts in Firefox (wrong positioning of background tiles). This pull request fixes this issue.


src/renderer/svg/svg.js
```diff
@@ -1219,27 +1220,29 @@ define([
...
-
-      pattern.setAttribute('width', 1 / fillRepeatX);
-      pattern.setAttribute('height', 1 / fillRepeatY);
+      pattern.setAttribute('patternUnits', 'userSpaceOnUse');
       pattern.setAttribute('x', 0);
       pattern.setAttribute('y', 0);
+      pattern.setAttribute('width', boundingBox.width / fillRepeatX);
+      pattern.setAttribute('height', boundingBox.height / fillRepeatY);
+
+      elementMatrix = attributes.matrix.clone();
```
padolsey, line 1229: Has this been tested in worker context? The real `Matrix` (with a clone method) instance will only be passed to the renderer in the iframe context ... in worker context, AFAIK, you just get a regular object.

klipstein, line 1229: @padolsey i'll try it in the worker now

klipstein, line 1229: TypeError: attributes.matrix.clone is not a function :(

basecode, line 1229: good point @padolsey. I totally overlooked that one.

klipstein, line 1229: @padolsey fixed in f3b7b86eb78b891971ae15fc04672a1a5d1141ee

PR #31, opened 2012-08-22
Remove existing `AssetController.preload`

basecode:


basecode:
The reason why I removed 
- AssetController.preload
- extensions.js

is because the former was broken and should be moved in a separate plugin and the latter is not used and doesn't give any useful informations.

More about extensions.js:

It was a collection of mime-types and corresponding AssetHandler. This map breaks as soon as a mime-type could be handled by more than one AssetHandler. For example `{ svg: 'bitmap', svg: 'fontFamily', svg: 'raw'}` . A possible script that needs to find out the correct AssetHandler for a certain mime-type would break at this point.

PR #39, opened 2012-08-28
Cleanup animation

basecode:
The reason behind removing setSubject and setSubjects is that they're probably not going to be used. addSubject(s)/removeSubject(s) are more than sufficient for what we foresee as common usages of the Animation class. The other changes are general code clean-up.

### DIFF
--- a/src/runner/animation/animation.js
+++ b/src/runner/animation/animation.js
@@ -19,6 +19,8 @@ define([
   'use strict';
 
   var mixin = tools.mixin;
+  var isArray = tools.isArray;
+  var forEach = tools.forEach;
 
   /**
    * Translations, in the form of:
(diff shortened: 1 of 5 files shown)


padolsey:
@basecode I think it's best if we have a short description for these types of pull requests. Someone in the future could look back and have no idea why these changes were made otherwise.

## 

I discussed these changes with Tobi. The reason behind removing setSubject and setSubjects is that they're probably not going to be used. addSubject(s)/removeSubject(s) are more than sufficient for what we foresee as common usages of the Animation class. The other changes are general code clean-up.

basecode:
You're right. A pull-request like this should have a description! And thank you for writing the description +2 :)

src/runner/animation/animation.js
```diff
@@ -257,21 +259,9 @@ define([
      *
      * Optionally changes the subject of the animation.
```
padolsey: I think this line needs to be removed.

PR #65, opened 2012-09-12
Pull out displaylist

davidaurelio:
- Changes the display list from mixin to composition
- `Group`, `Movie` and `Stage` now hold a `displayList` property
- User API remains unaffected – `addChild()`, `children()`, etc. still exist
- This enables custom display list behavior where needed by simply exchanging the implementation
- (de)registration has moved to `_activate()` / `_deactivate()`
- added 128 tests for display list behavior and lifecycle (many shared tests, the actual number is lower)
- started files for common/shared tests and mock objects


src/runner/display_list.js
```diff
@@ -219,87 +251,40 @@ define([
...
+      return DisplayObject.prototype.markUpdate.call(this);
+    },
+    removeChild: function(child) {
+      this.displayList.remove(child);
       return this;
     }
   };
 
-  return DisplayList;
+  var timelineMethods = tools.mixin({}, methods);
```
padolsey, line 290: Wondering if it's correct to have the timelineMethods defined in the display_list file?

padolsey:
The tests don't seem to have been added to runner.html. Not sure if they're intended to be build-tests only.

src/runner/display_list.js
davidaurelio, line 290: They could be moved to `movie.js`, since they’re only used there. Maybe it was too future-proof to make them available as mixin.

davidaurelio:
> The tests don't seem to have been added to runner.html. Not sure if they're intended to be build-tests only.

New files are
- test/common/displaylist-owner.js
- test/common/displayobject-lifecycle.js
- test/common/mock.js

these files contain shared tests and functionality and are included by other specs.

PR #66, opened 2012-09-14
Use first changed touch for the non-multi event

davidaurelio:
Using `domEvent.touches[0]` for the non-`'multi:*'` events fails for `'touchend'` and `'touchcancel'`, because said list is empty when e.g. releasing a single finger.

Changing to `domEvent.changedTouches[0]` solves this problem and is probably more accurate for `'touchstart'` as well.


padolsey:
Ah, makes sense. Thanks for the fix!

davidaurelio:
@qfox forced me to do it

Issue #51, opened 2012-09-04
Cleanup KeyframeAnimation class


basecode:
- `options.delay` is not supported
- setSubjects/setSubject could be removed
-  keyframes in seconds seems a little bit confusing

```
var rect = new Rect(0, 0, 200, 200).fill('green').addTo(stage);
var anim = new KeyframeAnimation('1s', {
  '5s': { x: 500 }
});
rect.animate(anim);
```

padolsey:
Not sure what to do about your last point, "keyframes in seconds seems a little bit confusing" ...

Currently you can define the full duration of the of the KeyframeAnimation in the first argument (s, ms, % etc.) and then you can define the individual keyframes using whatever units you want. I, for example, would normally use percent:

``` js
new bonsai.KeyframeAnimation('4s', {
    '0%': {x: 0, y: 0},
    '25%': {y: 0, x: 400},
    '50%': {x: 400, y: 400},
    '75%': {x: 0, y: 400},
    '100%': {x: 0, y: 0}
});
```

This is intuitive.

And I think it's okay to also have the option of defining the keyframes in seconds/milliseconds too. 

We can discuss removing that feature but we can keep it for "power users" and just not recommend it in the docs. Up for discussion.

basecode:
+1 I would also encourage users to use percent values.

basecode:
Current status:

> options.delay is not supported

This is a bug and still open.

> setSubjects/setSubject could be removed

removed in #75

> keyframes in seconds seems a little bit confusing

we keep them, but we encourage users to use percent values or `from/to`

padolsey:
Are we supporting `options.delay`? Where is that specified?

basecode:
@padolsey Meanwhile the KeyframeAnimation class supports `delay`, right?

padolsey:
Yep.

PR #75, opened 2012-09-18
KeyframeAnimation cleanup as per #51

padolsey:
See https://github.com/uxebu/bonsai/issues/51

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -2,6 +2,7 @@ v0.3.7
 -------------------
 * Fix issue where click event would be fired twice on touch devices (#60)
 * Change display list implementation from mixin to composition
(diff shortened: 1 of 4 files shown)


klipstein:
@padolsey are other projects affected by that change? i just remember that we once had to change it in `transwf`.

padolsey:
I've done a search for `setSubject` in transwf and can't find anything. I don't think transwf uses the KeyframeAnimation class either.

basecode:
Looks good! Could you pls add an entry to the change logs?

basecode:
Not sure about the tests. I think we already have tests that cover `addSubject`. Can't we just remove the tests regarding `setSubject`?

padolsey:
No new tests were added. I changed old tests that were using the setSubject(s) methods. Those specific tests were testing other KeyframeAnimation behaviour (so we should keep them).

Issue #77, opened 2012-09-19
`Make test` as a non uxebu-er


iamdustan:
I’m trying to dig deep into bonsai, but can’t run `make test` seeing as I don’t have access to https://github.com/uxebu/ci-tools from https://github.com/uxebu/bonsai/blob/master/Makefile#L14 :(

davidaurelio:
@iamdustan thanks for reporting. You are correct, the makefile is tailored to our CI server. We should make that a little friendlier.

To run the unit tests, you can also open [test/runner.html](https://github.com/uxebu/bonsai/blob/master/test/runner.html) in your browser to run unit tests. Unfortunately, the test coverage of components varies between rough and very good.

iamdustan:
Closed by #78

PR #78, opened 2012-09-19
Run jasmine tests in the console.

iamdustan:
Add phantomjs+jasmine console test runner with a new make command
`make test-local` to run tests quickly in the browser without the
need for uxebu's CI integration.

This would solve the issue of #77 with allowing any commiter to run
tests locally.


wolframkriesing:
this pull request is in the works
we might will come back to you asking for signing a CLA to a foundation, we are still in the process of figuring out how to do this as painless as possible, this is a topic many open source projects have fought with so we are sure the solution will be found soon, thanks

klipstein:
@iamdustan I've just checked the project https://github.com/jcarver989/phantom-jasmine and didn't found any license there. This pull request could be merged if phantom-jasmine would have a MIT, BSD or similar license or if this pull request would contain another project which provides a similar phantomjs-runner.

We definitely need to ensure that the copyright / IP is clear for BonsaiJS (see: http://www.outercurve.org/Blogs/EntryId/65/Open-Source-Software-Hygiene-and-STDs).

@jcarver989 would be great if you could add license information to phantom-jasmine

iamdustan:
@wolframkriesing sounds good. Completely understandable.

@klipstein agreed. Thanks for the link. Definitely going to read that. I’ll look around for some alternatives to the phantom/jasmine runners.

davidaurelio:
Just checked the runner on the command line and in the browser
- running on the command line is fast and works well
- running in the browser shows two HTML reporters and is ~3 times slower than before adding the console reporter.

Would it be possible to add the console reporter only for the phantom runner?

iamdustan:
@davidaurelio I accidentally added the `simpleReporter` a second time. Removing it brought the speed back up for me.

iamdustan:
Well, with the console closed it runs in the browser at ~0.8 seconds. With the console open it took ~1.2-1.5s. Does phantomjs have any dependable environment hooks so we can dynamically choose which reporter to hook into?

klipstein:
@iamdustan I think the only way is to check the user-agent:

``` js
var isPhantom = (navigator && navigator.userAgent.indexOf('PhantomJS') > -1);
```

iamdustan:
Thanks @klipstein. I was a little concerned on that since the docs here (http://code.google.com/p/phantomjs/wiki/Interface) say you can set the UA for phantomjs to run in.

Last commit has the consol running in ~0.5s and Chrome in ~0.7s.

Now just for that licensing...

iamdustan:
Theoretically Jasmine has a ConsoleReporter reporter itself, but I can’t seem to get it to work. It runs, but only prints bullets to the console and never exits.

https://github.com/pivotal/jasmine/blob/master/src/console/ConsoleReporter.js

klipstein:
@iamdustan using the user-agent for detecting should be fine for just executing the bonsai tests.

A general problem with PhantomJS is that you basically just can talk back from the browser to the executing script via "console.log" or by polling the PhantomJS browser.

The https://github.com/jcarver989/phantom-jasmine/blob/master/lib/run_jasmine_test.coffee#L36 script waits until it got "ConsoleReporter finished", which is not provided by the default ConsoleReporter. I assume it would exit, if you would check for "Finished in ..." (https://github.com/pivotal/jasmine/blob/master/src/console/ConsoleReporter.js#L88).

davidaurelio:
Thanks for adding the fork!

iamdustan:
Browsing the issues for phantom-jasmine I see he does reference what license it is released under. https://github.com/jcarver989/phantom-jasmine/issues/8

He says http://www.apache.org/licenses/LICENSE-2.0.html

Is that keeping it clean for bonsai or will a different project be required? If it is I’ll make a pull request for @jcarver989 with the license.

klipstein:
Apache 2.0 license is fine. Did you port the CS version (https://github.com/jcarver989/phantom-jasmine/blob/master/lib/run_jasmine_test.coffee) to JS? Wasn't able to find a reference to the JS version.

I would merge this pull request if you add a license note to https://github.com/uxebu/bonsai/blob/master/LICENSE on the bottom which links to the mentioned ticket. You probably should also rename `test/runner-jasmine.js` to `test/phantom-runner-jasmine.js` to make the purpose of this file more clear.

iamdustan:
I did. I just dropped it into http://js2coffee.org/. File renamed and license note added.

Would `make test-phantom` be a better name than `make test-local` or is that being a bit too pedantic?

klipstein:
I would say that `test-phantom` sounds more clear. `test-phantom` will probably become the default `make test` anyways.

To merge your pull and to make my live easier it would be great if you would merge it with latest master (test execution changed a little bit: https://github.com/uxebu/bonsai/pull/80).

iamdustan:
Done and done.

iamdustan:
In chrome: 831 specs, 0 failures in 0.711s
In phantom: 831 specs, 0 failures in 0.52s.

PR #88, opened 2012-09-27
Audio Class

basecode:
This pull request introduces the Audio Class, a new AssetDisplayObject. The Audio API provides attr-properties and methods to play, pause and stop a sound-file. Other than that you can control the volume and change the playblack position.


src/asset/asset_controller.js
```diff
@@ -9,16 +9,17 @@ define([
...
   './raw_handler'
 ],
 function(
   tools, EventEmitter, AssetRequest,
-  FontHandler, VideoHandler, BitmapHandler, RawHandler
+  FontHandler, VideoHandler, AudioHandler, BitmapHandler, RawHandler
 ) {
   'use strict';
 
-  // save references to all assets (TODO: rethink)
```
davidaurelio, line 21: Did you rethink it?

src/asset/asset_handler.js
```diff
@@ -5,8 +5,48 @@ define([
...
+      // in most cases the name of the handler is part of the description
+      var handler = (navMimeType.description.match(/audio|video|image/i) || [])[0];
+
+      // return early when the mime-type is irrelevant
+      if (!handler) {
+        return;
+      }
+
+      // case insensitive, support "video" and "Video"
+      handler = handler.toLowerCase();
```
davidaurelio, line 36: less complicated:

``` js
// in most cases the name of the handler is part of the description
var handler = navMimeType.description.match(/audio|video|image/i);

// return early when the mime-type is irrelevant
if (!handler) {
  return;
}

// case insensitive, support "video" and "Video"
handler = handler[0].toLowerCase();
```

src/asset/asset_controller.js
basecode, line 21: Yes, we have different strategies depending on the asset-type. Everything that is partially loaded (video, audio) is stored in `AssetController.assets`, which makes it accessibly by the AssetController (that triggers loading) and the renderer (that triggers the playback for example). For other asset-types that are accessibly via the browser's cache (assuming the file is cached) we don't need a shared context (access to the original Dom-Element) and therefore don't store a reference in `AssetController.assets`.

src/asset/asset_handler.js
```diff
@@ -5,8 +5,48 @@ define([
...
+
+      // store a mime-type per suffix (overwrites existing suffixes)
+      forEach(navMimeType.suffixes.split(','), function(suffix) {
+        if (suffix) {
+          mimeTypes[handler][suffix] = navMimeType.type;
+        }
+      });
+    });
+
+    return mimeTypes;
```
davidaurelio, line 46: Reducing might come in handy:

``` js
AssetHandler.MIME_TYPES = tools.reduce(
  typeof navigator !== 'undefined' ? navigator.mimeTypes : '', // '' is a zero-length iterable
  function(mimeTypes, navMimeType) {    // in most cases the name of the handler is part of the description
    var mediaType = navMimeType.description.match(/audio|video|image/i);

    // return early when the mime-type is irrelevant
    if (mediaType) {
      // case insensitive, support "video" and "Video"
      mediaType = mediaType[0].toLowerCase();

      // store a mime-type per suffix (overwrites existing suffixes)
      forEach(navMimeType.suffixes.split(',').filter(Boolean), function(suffix) {
        mimeTypes[mediaType][suffix] = navMimeType.type;
      });
    }

    return mimeTypes;
  },
  { audio: {}, video: {}, image: {} }
);
```

src/asset/asset_controller.js
davidaurelio, line 21: Sounds good.

Can we get rid of unloaded assets?

src/asset/audio_handler.js
```diff
@@ -0,0 +1,84 @@
...
+ */
+define([
+  './asset_handler'
+], function(AssetHandler) {
+  'use strict';
+
+  var AUDIO_MIME_TYPES = AssetHandler.MIME_TYPES.audio;
+
+  var domAudio = typeof document !== 'undefined' && document.createElement ?
+    document.createElement('audio') : 0;
```
davidaurelio: Save one byte by not introducing an arbitrary zero:

``` js
var domAudio = typeof document !== 'undefined' && document.createElement && document.createElement('audio');
```

Save even more bytes and make stuff more readable:

``` js
var domAudio;
try { 
  domAudio = document.createElement('audio');
} catch (e) {}
```

(this happens once, so the try/catch overhead does not matter)

src/asset/audio_handler.js
```diff
@@ -0,0 +1,84 @@
...
+    'loadeddata': 'loadeddata',
+    'canplay': 'canplay',
+    'canplaythrough': 'canplaythrough'
+  };
+
+  function AudioHandler() {
+    AssetHandler.apply(this, arguments);
+  }
+
+  var playableMimeType = AudioHandler.playableMimeType = function(mimeType) {
```
davidaurelio: Why no `canPlayType` as in the original?

src/renderer/svg/svg.js
```diff
@@ -76,7 +79,7 @@ define([
 
   function createElement(n, id) {
     if (!n) {
-      throw TypeError('Invalid tag name: ' + n);
+      throw new TypeError('Invalid tag name: ' + n);
     }
```
davidaurelio, line 83: https://github.com/uxebu/guides/blob/master/coding-style.md#errors

src/renderer/svg/svg.js
```diff
@@ -651,16 +650,66 @@ define([
...
+    var volume;
+    var attributes = message.attributes;
+    var id = message.id;
+    var playing = attributes.playing;
+
+    if (typeof audioElement === 'undefined') {
+      throw new Error('asset <' + id + '> is unknown.');
+    }
+
+    if (attributes.prepareUserEvent) {
```
davidaurelio: Would `&& 'ontouchstart' in audioElement` make sense to avoid adding handlers that will never be removed again?

src/renderer/svg/svg.js
```diff
@@ -651,16 +650,66 @@ define([
...
+      // We bind to the next touch-event and play/pause the audio to cause
+      // iOS devices to allow subsequent play/pause commands on the audio el.
+      // --
+      // (Usually, iOS Devices will only allow play/pause methods to be called
+      // after a user event. Due to bonsai's async nature, a movie programmer
+      // can never achieve this. So we setup a fake one here...)
+      var touchStartHandler = function() {
+        audioElement.play();
+        audioElement.pause();
+        document.removeEventListener('touchstart', touchStartHandler, true);
```
davidaurelio, line 674: Should we send a `'canPlay'` event? How are native events affected by this?

davidaurelio, line 674: What if audio should actually be playing at this point? Is that handled elsewhere?

src/renderer/svg/svg.js
```diff
@@ -651,16 +650,66 @@ define([
...
+
+    if ('volume' in attributes) {
+      // Value between 0-1. NaN is treated as `0`
+      audioElement.volume = min(max(+attributes.volume || 0, 0), 1);
+    }
+
+    // Time in seconds. `currentTime` throws when there's no
+    // current playback state machine
+    if ('time' in attributes) {
+      // Set volume to 0 to avoid "clicks"
```
davidaurelio, line 687: thumbs up. very nice you’ve considered this

src/runner/audio.js
```diff
@@ -0,0 +1,182 @@
...
+    });
+  };
+
+  /** 
+   * Prepare the Audio object for a user-event. 
+   * (currently this is for iOS devices, see drawAudio method in svg.js)
+   *
+   * @returns {Audio} this
+   */
+  proto.prepareUserEvent = function() {
```
davidaurelio, line 154: Hm, I’d love to come up with a better method name, but right now I can’t.

davidaurelio:
General picture is good, I’ve added a couple of remarks and questions

src/runner/audio.js
padolsey, line 154: It's horrible because we're letting browser inconsistencies leak into our API. There's no easy way to avoid this.

src/renderer/svg/svg.js
padolsey: > never be removed again

They would seem to get removed in `removeEventListener` below. I see your point though, there's no point in binding this for non-touch devices.

src/runner/audio.js
davidaurelio, line 154: should we simply do this by default? There isn’t any harm, is there?

src/renderer/svg/svg.js
davidaurelio: Yes, I should have been more clear. It won’t ever be removed on non-touch devices, effectively leaking the `<audio>` element.

src/runner/audio.js
padolsey, line 154: It could mean that we're forcing the browser to load stuff that it doesn't need to yet.

I guess it'd make things simpler though. We could add this to the Audio constructor, for example:

``` js
this.on('addedToStage', this.prepareUserEvent);
```

davidaurelio, line 154: Ok, looks like a good idea. No need to do it right now, I will also pull without changes here.

src/asset/asset_controller.js
basecode, line 21: Affected AssetDisplayObjects overwrite `destroy()` and remove the reference.

src/renderer/svg/svg.js
padolsey, line 674: For touch devices, audio simply cannot play without initially calling calling `prepareUserEvent`. I'm not sure of a case where it'll already be playing...

davidaurelio, line 674: I meant: What if the user had actually already set the bonsai audio object to playing? The `<audio>` element is not playing yet, of course.

padolsey, line 674: If the user sets the Audio instance to play before `prepareUserEvent` is called then the associated HTML5 audio element will not play. This touchstart event handler we add for touch-devices will cause the audio element to play then immediately pause (<- this is the broken? bit). If the programmer then calls `play` on the audio instance it will play correctly.

I'm not sure what the expected behaviour should be though. The touchstart event will respond to anything on the document. I'm sure the programmer doesn't want their audio to start playing at a random place. In the scenario when they called `play()` they would expect it to play immediately. If they haven't yet called `prepareUserEvent()` then we cannot achieve this anyway because of the iOS limitation. It's best, I think, for us to just fail (not play) and then work when they next call `play()` (after `prepareUserEvent` has done its stuff), rather than start playing as soon as some arbitrary touchstart event occurs.

src/asset/asset_handler.js
basecode, line 36: Regarding readability? Because I wanted `handler` to be a string and not an array. In contrast to an array, a string indicates _one_ result.

src/asset/audio_handler.js
padolsey: https://github.com/uxebu/bonsai/commit/28176b18461112e7101731f1472b15aa2151e0be

src/asset/asset_handler.js
davidaurelio, line 46: @basecode just told me you guys are changing stuff. Please don’t feel forced to do this. I just wanted to advertise the services of my good friend `reduce` ;-)

src/renderer/svg/svg.js
basecode, line 83: fixed

src/asset/audio_handler.js
basecode: Because it isn't the same as "canPlayType". It returns a mimetype. Renamed to "getPlayableMimeType". Fixed. Thanks!

src/asset/audio_handler.js
basecode: Good idea. Fixed

src/runner/audio.js
basecode, line 154: Maybe we'll find a better solution. I added it to a list of open issues. See: https://github.com/uxebu/bonsai/issues/92

basecode:
@davidaurelio, do you think we're ready to merge?

davidaurelio:
absolutely

PR #89, opened 2012-09-27
Set margin/padding to 0 for the body of foreignobject.

klipstein:
Fixes issues where default margin / padding is applied on textfields.


klipstein:
We will solve this issue by rendering the result in an iframe.

Issue #79, opened 2012-09-20
Event passing between page context and movies


iamdustan:
@davidaurelio I’m told your the man for this.

Currently, the only way to pass events inside a movie back to the parent page and vice versa is through the `sendMessage` method. Behind the scenes I presume this will be the path things will need to travel through (for iframe/webworker communication).

Example: https://gist.github.com/3756383

Having to pass user events back and forth this way is suboptimal though. It would be much more intuitive to just emit events and have it captured at each level.

Example: https://gist.github.com/744cf037755422eb888c

davidaurelio:
Thank you for the input. I can see your point and agree that the current situation is not ideal.

For a couple of reasons, I hesitate to connect _main page_ and _stage_ (within the bonsai environment) to a universal emitter.
- I have the impression that if users emit an event on either of the two, they usually intend to communicate something to the other side. A connected event emitter would create the need to differentiate between senders. Say, somebody wants to send `'message'` events from _main page_ to _stage_ and vice-versa, listeners on both sides would need to differentiate. This would probably happen by using two different event types, e.g. `'stage-message'` and `'page-message'`.
- While there are not many events emitted on the _stage,_ four of them could cause significant overhead if also sent to the _main page_: 
  - `'exitFrame'` is fired for each frame and probably of little interest on the _main page_ in many cases. 
  - `'pointermove'` (i.e. `'mousemove'` and `'touchmove'` events) are already piped from _main page_ to the _stage_ and would need to be sent back.
  - `'render'` is emitted for every render cycle (not necessarily the same as a frame)
  - `'submoviesAdvanced'` is emitted in every frame when all sub movies are advanced.

This might be trivial locally (not sure), but we also want to support a remote node server running a bonsai movie.

We still might consider doing this, but we’ll need some time to benchmark and evaluate the results.

If I’ve understood you correctly, the most tedious thing about `sendMessage(data)` is the fact that it is uncategorized, i.e. your message listeners need to branch according to message types. We have the same impression and were already considering to change it to `sendMessage(type, data)`, which would emit a `'message:<type>'` event on the other side.

This is not meant to be implemented _instead_ of your “universal emitter” idea, but could ease the pain for the short term.

Your example would become:

``` js
// main.js
var _stage = document.getElmenetById('stage'),
    stage = bonsai.run(_stage, 'movie.js');

stage.on('message:pointer-up-on-red-square', function () {
  console.log('user-event', arguments);
})
```

``` js
// movie.js
[...]

PR #95, opened 2012-09-29
Categorized messages

davidaurelio:
- Is backwards compatible to the current signature of `sendMessage()`
- Adds optional `category` parameter to `sendMessage`
- categorized messages trigger a `'message:<category>'` event
- is the “small solution” described in #79
- should make uxebu/bonsai-docs#20 easier

### DIFF
--- a/src/runner/stage.js
+++ b/src/runner/stage.js
@@ -149,11 +149,8 @@ define([
           this.env.exports.env.emit('change', data);
           break;
         case 'message':
-          var category = message.category;
-          if (category) {
-            this
-              .emit('message:' + category, data)
-              .emit('message', data, category);
+          if ('category' in message) {
+            this.emit('message:' + message.category, data);
           } else {
             this.emit('message', data);
           }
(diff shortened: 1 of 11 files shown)


src/runner/stage.js
```diff
@@ -149,7 +149,14 @@ define([
           this.env.exports.env.emit('change', data);
           break;
         case 'message':
-          this.emit('message', data);
+          var category = message.category;
+          if (category) {
+            this
+              .emit('message:' + category, data)
+              .emit('message', data, category);
```
iamdustan: Is this emitting multiple messages so that both explicitly namespaced events and the generic handlers can capture it?

davidaurelio: Yes, exactly. I thought it might come in handy. Does it look like a good or bad idea to you?

I intentionally kept the invocation of the callback for uncategorized messages as is (with one argument), but the more I think about it, the addition of categories might interfere with the old approach anyways.

Maybe the whole block could be simplified to

``` js
case 'message':
  var category = message.category;
  if (category) this.emit('message:' + category, data);
  this.emit('message', data, category);
  break;
```

iamdustan: I see where you’re coming from with including it.  As to good or bad—my primary concern is that it could cause unwanted side effects by triggering an event handler I didn’t mean to trigger.

Does the current implementation of the event emitter support namespaced events? If so, then would the second emit be covered automatically by the namespaced event and the user could have a chance to filter out the event based on that? Otherwise, this might cause unwanted side effects which would be trickier to overcome than just letting the user emit both events in their code.

davidaurelio: Ok, I’ve changed it: Messages with a category will only invoke namespaced listeners.

The event emitter itself has no special support for namespaces

davidaurelio:
@basecode @padolsey could one of you review, please?

PR #97, opened 2012-10-02
Providing just one build-type (worker+iframe)

klipstein:
This pull-request removes the need to provide 2 different builds for the Worker / Iframe environment and it allows that Bonsai can be concatted with other JS files (the complete bonsai-code is now injected through different techniques in the worker/iframe environment as a string instead of a file URL).

The built version of Bonsai is now taking care to select the best approach for a particular browser:
- Worker-Environment spawned with Blob (latest Chrome, Firefox, Safari 6 and iOS 6)
- Worker-Environment spawned with Data-URI (Opera 12)
- Iframe-Environment spawned through `<script>/*bonsaiCode*/</script>` (older Chrome versions, Safari 5, IE 9/10. Note: IE10 throws a security error, if a Blob-URL is used as the Worker source)

To just allow the bonsai-code to be executed in the iframe context you can do the following:

``` js
bonsai.setup({
  runnerContext: bonsai.IframeRunnerContext
}).run(stage, 'movie.js');
```

Or if you want to limit it to the worker context which loads bonsai through a URL you can do the following:

``` js
bonsai.setup({
  runnerContext: bonsai.WorkerRunnerContext,
  runnerUrl: 'http://url/to/bonsai.js'
}).run(stage, 'movie.js');
```


iamdustan:
+1.

I think this would help with the confusion those two fellas were talking about in IRC this past weekend (September 29).

PR #106, opened 2012-10-04
Animation Refactor

padolsey:
This pull request includes various changes to the animation APIs in bonsai.

The aim was to simplify and generalise the Animation APIs.
#### Animation class destroyed

The Animation class now only exists as a convenience factory for `KeyframeAnimation`.

`new Animation` now returns a single-phase `KeyframeAnimation` instance. So, animations are really just keyframe animations now. 
#### Strategies removed

Previously we supported strategies such as `prop`, `attr` or custom strategies supplying `get` and `set` methods (these were used by the Animation to apply animated values to subjects). The support for strategies has now been **removed**. 

`KeyframeAnimation` now supports a single way of setting and getting properties from subjects: the `attr` method. This is our main use-case (animating DisplayObject) and if you need to you can still create any arbitrary animation as long as you provide the `attr` method. See [this example](https://gist.github.com/3835790).
#### New `PropertyTween` class

This takes care of tweening a single property like `x` or `fillColor` or `width`. The `PropertyTween` applies the  _optional_ passed translator to the values. A translator is used to split a property value into its component parts (e.g. `fillColor` -> `r,g,b,a`) and then back to its true type on each `at()` call (after actual tweening).

PropertyTween example:

``` js
var p = new PropertyTween(0, 50);
p.at(0); // => 0
p.at(0.5); // => 25
p.at(0.8); // => 40
p.at(1); // => 50
```

With a translator:

``` js
var colorTranslator = {
  toNumeric: function(c) {
    return [c.r(), c.g(), c.b(), c.a()];
  },
  toUnique: function(n) {
    return new color.RGBAColor(n[0], n[1], n[2], n[3]);
  }
};
var tween = new PropertyTween(color('red'), color('blue'), colorTranslator);

tween.at(0); // RGBAColor instance
tween.at(0).rgba();   // => "rgba(255,0,0,1)"   (red)
tween.at(0.5).rgba(); // => "rgba(128,0,128,1)" (purple)
tween.at(1).rgba();   // => "rgba(0,0,255,1)"   (blue)
```
#### New `PropertiesTween` class

This takes care of tweening multiple properties. To do this it instantiates a
`PropertyTween` instance for each individual property.

The `PropertiesTween#at` method will return a whole object containing all values
at the passed progress. E.g.

``` js
var p = new PropertiesTween({x: 0}, {x: 1000});

p.at(0).x; // => 0
p.at(0.1).x; // => 100
p.at(0.5).x; // => 500
```

It also handles all of the default translators we have. Currently we have 
translators for these properties:
- fillColor
- strokeColor
- fillGradient
- strokeGradient
- matrix
- filters
- segments [used for morphing]
- cornerRadius [array of radiuses for Rect]

So, if I pass any of those properties, translation will be done automatically.
E.g.

``` js
var p = new PropertiesTween(
  { matrix: new Matrix(1,0,0,1,0,0) },
  { matrix: new Matrix(2,0,0,1,200,200) }
);

p.at(0.5).matrix; // => new Matrix(1.5,0,0,1,100,100)
```
#### Translators are more efficient

Translators have a simple contract. They must provide:
- A `toNumeric` method: _Is passed a raw start or end value and returns an array of numbers that can be tweened._
- A `toUnique` method: _Is passed an array of numbers and returns a value that can be returned by `PropertyTween#at()`. This is usually the custom value like a color or matrix instance which will be composed from the tweened numbers in the array._

Previously translators had to build, maintain and destroy named properties in custom `values` objects ... Now it's MUCH simpler; they just need to provide the above methods, and they're more performant as they're just dealing with arrays. Compare the [old](https://github.com/uxebu/bonsai/blob/master/src/runner/animation/matrix_translations.js) with the [new](https://github.com/uxebu/bonsai/blob/animation-generic/src/runner/animation/translators/matrix.js)!
#### Notes:
- It might be a good idea to expose `PropertiesTween` and `PropertyTween` classes. I imagine they could be useful as stand-alone functionality. Not sure though.
- All example movies work. All tests work. New tests have been added for new endpoints/classes.


example/library/index.html
```diff
@@ -77,6 +77,7 @@
         url: 'movies/' + script + '?' + new Date().getTime(),
         width: document.width,
         height: document.height - 50,
+        framerate: 100,
         fpsLog: /animation/.test(script) // only show FPS log on animations
```
basecode, line 81: Are you sure `100` makes sense? We set Orbit to `60` for example.

src/runner/animation/keyframe_animation.js
```diff
@@ -91,143 +143,131 @@ define([
...
-     *   - 'attr': The 'attr' method of the object is used (for DisplayObjects)
-     *   - 'prop': Normal property setting and getting is used
-     *   - Object with 'set(subject, values)' and 'get(subject)'
-     *     methods.
      */
-    addSubjects: function(subjects, strategy) {
+    addSubjects: function(subjects) {
       var me = this;
       subjects = tools.isArray(subjects) ? subjects : [subjects];
       subjects.forEach(function(subject) {
```
basecode: Would you mind to replace all `[].forEach` with `tools.forEach` please?

src/runner/animation/keyframe_animation.js
```diff
@@ -1,9 +1,9 @@
...
   './easing',
   '../../tools',
-  '../../event_emitter'
-], function(Animation, easing, tools, EventEmitter) {
+  '../../event_emitter',
+  './properties_tween'
+], function(easing, tools, EventEmitter, PropertiesTween) {
   'use strict';
 
   var max = Math.max,
```
basecode: I think you can remove `max` and `round`. They seem not to be used anymore.

src/runner/animation/keyframe_animation.js
```diff
@@ -341,7 +358,7 @@ define([
       });
 
       // Fill in (missing) properties:
-      keys.forEach(function(frame, i) {
+      tools.forEach(keys, function(frame, i) {
```
basecode: I guess it makes sense to cache the property. `var forEach = tools.forEach;`

Maybe also:
`var forOwn = tools.forOwn;` and remove all occurrences of `for…in`.

src/runner/animation/properties_tween.js
```diff
@@ -0,0 +1,114 @@
...
+    _setupTweens: function() {
+
+      var easingFn = this.easingFn,
+          propertiesFrom = this.propertiesFrom,
+          propertiesTo = this.propertiesTo,
+          propertyNames = this.propertyNames,
+          propertyTweens = this.propertyTweens;
+
+      for (var i = 0, l = this.propertyLength; i < l; ++i) {
+        var property = propertyNames[i];
```
basecode: To avoid creating pseudo-scoped variables like `var property` that leaks into the parent-scope I would suggest to use tools.forEach here. `tools.forEach(propertyNames, function(property) { …  })`.

src/runner/animation/property_tween.js
```diff
@@ -0,0 +1,59 @@
...
+    this.translator = translator;
+  }
+
+  /**
+   * Returns the value at the passed progress (translator is applied)
+   *
+   * @param progress
+   */
+  PropertyTween.prototype.at = function(progress) {
+
```
basecode, line 42: I like it!

src/runner/animation/property_tween.js
```diff
@@ -0,0 +1,59 @@
...
+    var numericStartValues = this.numericStartValues,
+        numericEndValues = this.numericEndValues,
+        numericCurrentValues = [];
+
+    for (var i = 0, l = this.length; i < l; ++i) {
+      var startValue = numericStartValues[i],
```
basecode, line 48: See: pseudo-scoped vars

src/runner/animation/animation.js
```diff
@@ -77,436 +33,13 @@ define([
...
-
-    this._parseEventProps(options);
+  function Animation(clock, duration, endValues, options) {
 
-    var easingFunc = options.easing;
-    this.easing = typeof easingFunc == 'function' ?
-      easingFunc : easing[easingFunc];
+    return new KeyframeAnimation(clock, duration, {
+      to: endValues
+    }, options);
```
davidaurelio, line 40: :-) Makes me smile. I like

src/runner/animation/keyframe_animation.js
davidaurelio: That’s right, and there’s more. I can go through it in my IDE after the merge

src/runner/animation/keyframe_animation.js
```diff
@@ -37,52 +35,106 @@ define([
...
 
+    this._parseEventProps(options);
+
     this.subjects = [];
-    this.animations = [];
     this.initialValues = null;
-    this.currentAnimation = -1;
 
-    this.repeat = options.repeat || 0;
+    this.repeat = (options.repeat || 0) - (options.repeat % 1 || 0);
```
davidaurelio, line 47: Could be shortened to

``` js
this.repeat = options.repeat - options.repeat % 1 || 0;
```

src/runner/animation/keyframe_animation.js
```diff
@@ -91,143 +143,131 @@ define([
...
+      this._step(frame / duration);
+
+      if (
+        (this.isTimelineBound && timelineIsFinished) ||
+        frame === duration
+      ) {
+        this.prevFrame = 0;
+        this.currentDelay = this.delay;
+        this.reset();
+        if (this.repeat === Infinity || this.repeat-- > 0) {
```
davidaurelio: No need to check for `Infinity` explicitely. `Infinity - 1` remains infinite.

src/runner/animation/keyframe_animation.js
```diff
@@ -91,143 +143,131 @@ define([
...
+          this.play();
+        } else {
+          this.emit('end', this);
         }
```
davidaurelio, line 202: This block looks a little confusing to me. I could understand what it does after looking at the `play` and `reset` methods, but I think it’d be much clearer if it looked something like this:  

``` js
this.rewind();
this.repeat -= 1;
if (this.repeat < 0) {
  this.reset();
  this.emit('end', this);
}
```

with `rewind()` and `reset()` like

``` js
rewind: function() {
  this.frame = this.currentTweenIndex = this.prevFrame = 0;
  this.currentDelay = this.delay;
  return this;
},

reset: function() {
  this.rewind();
  this.isPlaying = false;
  this.clock.removeListener(this.isTimelineBound ? 'advance' : 'tick', this, this._onStep);
  return this;
}
```

`reset`/`rewind` could then also be used in the constructor to remove some duplication

src/runner/animation/keyframe_animation.js
davidaurelio: I wouldn’t replace regular loops in hot spots if they don’t consist of a single function call, but rather contain inline code. Stuff will be running every frame, so reducing overhead is a good thing in those spots, even if not as elegant.

padolsey: All the places where forEach is used aren't hotspots. I mean, they're not run every frame -- all of them are setup functions, used when initialising an animation or adding a subject.

src/runner/animation/properties_tween.js
padolsey: I'll change this to a forEach call, just for consistency, but I really don't see a problem with "psuedo" scoped vars. As long as the programmer is aware that we only have function (not block) scope then I don't see the problem. I guess in this case you could move the declaration of the `property` var up to the top of the fn..

src/runner/animation/keyframe_animation.js
padolsey, line 47: Cool thanks. I'll add this.

src/runner/animation/keyframe_animation.js
padolsey: Ah, good point!

src/runner/animation/keyframe_animation.js
padolsey, line 202: Hmm, I'm not sure. I think there is still ambiguity here. At first glance I am not sure of the differences between `reset()` and `rewind()`. 

Also, what's wrong with `if (this.repeat-- > 0) {` ? :( I really like it.. I guess it's not very intuitive though.

src/runner/animation/keyframe_animation.js
davidaurelio: I’ve added this comment where it does not make any sense. Just wanted to say I’d not necessarily replace every imperative loop with a functional one

src/runner/animation/keyframe_animation.js
davidaurelio, line 202: There’s nothing wrong with `this.repeat-- > 0` :-) I changed it while rearranging, b/c it matches my taste better if an expression doesn’t do two things at the same time, but leave it as is.

src/runner/animation/keyframe_animation.js
padolsey: I agree. I don't want to use functional loops everywhere, but I do want consistency.

src/runner/animation/keyframe_animation.js
davidaurelio, line 202: You’re right, `rewind()`/`reset()` don’t really help to reduce ambiguity.

My original motivation to comment on this was: I found calling `reset()` and then `play()` looks a little confusing, and also adds some overhead by removing and re-registering the `tick` handler.

src/runner/animation/property_tween.js
```diff
@@ -0,0 +1,59 @@
...
+   * @param {Object} [translator] The translator to be applied to values
+   * @param {Function} [translator.toNumeric] Is passed a raw start or end
+   *  value and returns an array of numbers that can be tweened.
+   * @param {Function} [translator.toUnique] Is passed an array of numbers and
+   *  returns a value that can be returned by `at()`. This is usually the 
+   *  custom value like a color or matrix instance which will be composed from
+   *  the tweened numbers in the array.
+   */
+  function PropertyTween(startValue, endValue, translator) {
+    this.translationData = {};
```
basecode, line 28: What is `this.translationData` good for? Maybe you could add a comment?

src/runner/animation/property_tween.js
```diff
@@ -0,0 +1,59 @@
+define([
+  './easing'
+], function(easing) {
+
```
basecode, line 2: `easing` is not used

src/runner/animation/property_tween.js
```diff
@@ -0,0 +1,59 @@
...
+   *
+   * @constructor
+   * @name PropertyTween
+   * @private
+   * @param {*} startValue Starting value of the tween
+   * @param {*} endValue Ending value of the tween
+   * @param {Object} [translator] The translator to be applied to values
+   * @param {Function} [translator.toNumeric] Is passed a raw start or end
+   *  value and returns an array of numbers that can be tweened.
+   * @param {Function} [translator.toUnique] Is passed an array of numbers and
```
basecode: Do you think the name `toUnique` describes what the function does? It sounds it actually returns sth that could be also returned by `attr`. Would it make sense to rename it to `toAttr`? Maybe I missed something.

src/runner/animation/keyframe_animation.js
```diff
@@ -60,29 +70,78 @@ define([
...
+      this.emit('play', this);
+
+      /*
+        Handle the case where initial values are specified and are
+        different from the subject's values. We need to set these
+        `from` properties manually: (FRAME 0)
+      */
+      var initial = this.keyframes[0];
+      if (initial && this.currentTweenIndex === 0) {
+        var subjects = this.subjects;
```
basecode: `subjects` is already defined. See pseudo-scope vars ;)

src/runner/animation/property_tween.js
padolsey: Good point. I'll change it to `toAttr`.

Issue #2, opened 2012-07-10
getComputed('width') doesn't respect borders


wolframkriesing:
The following always alerts '100' no matter what value the `lineWidth` has

``` js
var s = Shape.rect(100,100,100,100)
  .addTo(stage)
  .attr({fillColor:'red', lineWidth:10, lineColor: 'green'});
alert(s.getComputed('width'))
```

padolsey:
Currently the `getComputed` method determines the bounding box using just the paths. To include the borders too we'd need to also add considerations for `join` method used for lines (i.e. miter, round, bevel). I'll investigate this further.

Issue #116, opened 2012-10-11
getComputed() wrong with negative values


wolframkriesing:
The following 

``` js
var rect = new Rect(100, 100, -100, -100)
  .attr({fillColor: 'red'})
  .addTo(stage);

console.log(rect.attr('x'));
console.log(rect.attr('y'));
console.log(rect.getComputed('width'));
console.log(rect.getComputed('height'));
```

prints four times, 100, though x+y should be 0

iamdustan:
Wouldn’t that end up being a Rectangle that just had it’s origin point in the visible bottom right corner? I would expect the bounding box of that element to be `x1: 0, y1: 0, x2: 100, y2: 100`? I don’t know if I would expect the getComputed('width') to be `100` or `-100`.

wolframkriesing:
yep, that would make most sense but might have influence on any further
operations, since the origin was moved
i actuakly thought the computed width and height should -100
i let u guys figure out, for my use case both works  but thats not what
counts, lets find the right solution!
Am 12.10.2012 02:26 schrieb "Dustan Kasten" notifications@github.com:

davidaurelio:
> though x+y should be 0

No, they should equal the values you’ve passed to the constructor, i.e. `100`. The signature of `Rect` is `x, y, width, height`.

The _computed_ left/top values should be 0, and that’s what `getComputed` returns for them

``` js
>>> rect.getComputed('size');
{ top: 0, right: 100, bottom: 100, left: 0, width: 100, height: 100 }
```

> I don’t know if I would expect the getComputed('width') to be 100 or -100

We are going with `100`, because it felt more intuitive to return the visible size. Please reopen if this feels wrong.

wolframkriesing:
I understand the width/height=100, but that makes the x,y,width,height tuple useless, and that was the trap i stepped in. didnt even know about "size". if we wanna stick to it we might get some clearer naming ... dont know how, but the current solution is very irritating

iamdustan:
@wolframkriesing agreed that the current `getComputed` is a tricky road to travel at present. Does the discussion at #74 address this as well?

davidaurelio:
- Yes, “size” is not good, see #74.
- Yes, the tuple `[attr('x'), attr('y'), getComputed('width'), getComputed('height')]` tuple is not really useful; should we document that explicitly?

davidaurelio:
[...]

Issue #74, opened 2012-09-17
getComputed


basecode:
I was missing a method to get the bounding box of a DisplayObject. I saw that we're supporting `getComputed` instead, which has a even richer API and can return e.g. the "right" or "bottom" offset. But it still doesn't return the bounding box, which is supposed to return an object like `{x, y, width, height}`.
@davidaurelio proposed a new paramater `box` that can be passed to `getComputed`. I would love to field 2 other parameter names. See the list. Would do you like best?
- `getComputed('box')`
- `getComputed('bbox')`
- `getComputed('boundingbox')`

padolsey:
I thought that's what `getComputed('size')` does?

padolsey:
Here: https://github.com/uxebu/bonsai/blob/master/src/runner/path/path.js#L1125-1130

davidaurelio:
Yes, `'size'` does it. My proposal was to alias it with `'box'` if that’s more familiar

basecode:
I think `'size'` is supposed to return `{width:x, height:x}` and not the offset, position or bounding box.. Just the size.

basecode:
I propose the following return values:

`getComputed('size')` => `{width:0, height:0}`
`getComputed('position')` => `{x:0, y:0}`
`getComputed('box')` => `{x:0, y:0, width:0, height:0}`

Maybe also `right`, `bottom`. I would remove `left` and `top` b/c we already have `x` and `y`, which is also known from `attr()`.

klipstein:
I like @basecode's proposal, but would not add `right` and `bottom`, because this could be calculated easily from `x`/`y`/`width`/`height`.

davidaurelio:
- width and height are actually calculated from left and right, top and bottom: https://github.com/uxebu/bonsai/blob/master/src/runner/path/path.js#L1144-1145
- “left” and “top” are named as such to stress their relation to “bottom” and “right”

While I am totally open to discuss misleading naming (e.g. rename “size” to “box” or similar), I’m against making things more complicated and provide less information that we can, just because the users could calculate it themselves.

What’s the point in having separate “size” and “position” keys if “box” would give users everything? Why leave things out we can calculate easily?

basecode:
> While I am totally open to discuss misleading naming

Cool, because "top" is not a "size".

> What’s the point in having separate “size” and “position” keys if “box” would give users everything?

Okay, I'm fine with "box".

> Why leave things out we can calculate easily?

[...]

PR #121, opened 2012-10-18
getBoundingBox improvements

padolsey:
What was previously the `getComputed` method is now `getBoundingBox`.

It takes on optional argument, a `Matrix` instance which will be used to apply transforms to the calculated bounding-box. 

``` js
var shape = new Rect(50, 50, 100, 100);

// No transform applied, -- it'll return the bounding-box
// within shape's own coordinate system:
shape.getBoundingBox(); // ~ 0,0,100,100

// Apply shape's own matrix to the calculation
// (gives position in outer coordinate system):
shape.getBoundingBox( shape.attr('matrix') ); // ~ 50,50,100,100
```

`getBoundingBox` always returns an object in the form:

``` js
{
  top: n,
  left: n,
  width: n,
  height: n,
  bottom: n,
  right: n
}
```

Included in this pull request is also an algorithm to determine the correct extrema of cubic bezier curves. Checkout the branch and open the `bounding-box.js` example to see a demonstration of the new `getBoundingBox` calculation.


basecode:
- Afaik we agreed on `x` and `y` as aliases for top and left.
- API wise I think we should think about a convenient wrapper for `shape.getBoundingBox( shape.attr('matrix') );`
  maybe shape.getBoundingBox(true);  => I know `true` says nothing, but you get the point.

src/runner/path/curved_path.js
```diff
@@ -570,6 +570,51 @@ define([
...
+    var p0 = [p0x, p0y];
+    var p1 = [cp1x, cp1y];
+    var p2 = [cp2x, cp2y];
+    var p3 = [p1x, p1y];
+    var bounds = [[], []];
+
+    bounds[0].push(p3[0]);
+    bounds[1].push(p3[1]);
+
+    for (var i = 0; i < 2; ++i) {
```
basecode, line 591: Could you add a URL that points to the origin of this approach and if possible add some comments?

src/runner/path/curved_path.js
```diff
@@ -570,6 +570,51 @@ define([
...
+      var b = 6 * p0[i] - 12 * p1[i] + 6 * p2[i];
+      var a = -3 * p0[i] + 9 * p1[i] - 9 * p2[i] + 3 * p3[i];
+      var c = 3 * p1[i] - 3 * p0[i];
+      if (a == 0) {
+        if (b == 0) continue;
+        var t = -c / b;
+        if (0 < t && t < 1) bounds[i].push(f(t))
```
basecode: semicolon and brackets are missing

src/runner/path/curved_path.js
padolsey, line 591: This is the worst part of the implementation because I am not sure how it works. I ported it from a python script I found here: http://blog.hackers-cafe.net/2009/06/how-to-calculate-bezier-curves-bounding.html

I tried understanding the math on the article you linked to but in the end, after struggling, I grabbed this and ported it and it seemed to work.

iamdustan:
For related issues and discussions:
- #74 — getComputed
- #2 — getComputed('width') doesn't respect borders
- #116 — getComputed() wrong with negative values

These should all be closable when this is merged.

src/runner/path/curved_path.js
klipstein, line 591: @padolsey please add his MIT License note to our LICENSE file.

src/runner/display_list.js
```diff
@@ -229,53 +229,49 @@ define([
...
-    getComputed: function(key) {
+    /**
+     * Computes bounding boxes and single data points of a display object.
+     *
+     * @param {String} key What to compute. One of "top", "right", "bottom",
+     *    "left", "width" and "height"
+     * @param {Matrix} [transform=null] A transform to apply to all points
+     *    before computation.
+     * @return {Object}
+     */
```
davidaurelio, line 238: The doc block still has the `key` parameter

src/runner/display_list.js
```diff
@@ -229,53 +229,49 @@ define([
...
+    getBoundingBox: function(transform) {
+
       var children = this.displayList.children;
```
davidaurelio, line 241: if there are no children, the `reduce` call will have no effect. In that case, the method will always return `top: 0, right: 0, bottom: 0, left: 0, width: 0, height: 0`.

The code for empty objects already lives in `DisplayObject.prototype.getBoundingBox()`.

Adding a guard here should do the trick.

``` js
if (!children.length) {
  return DisplayObject.prototype.getBoundingBox.call(this, transform);
}
```

And a unit test for this special case might be handy

src/runner/path/path.js
```diff
@@ -140,6 +142,8 @@ define([
     this._owner.markUpdate('shapeData');
   }
 
+  Path.CurvedPath = CurvedPath;
```
davidaurelio: This does not seem to be accessed from anywhere

src/runner/video.js
```diff
@@ -82,23 +82,19 @@ define([
...
+    var box = {top: 0, right: 0, bottom: 0, left: 0};
+    box.right = box.width = this.attr('width') || 0;
+    box.bottom = box.height = this.attr('height') || 0;
+    if (transform) {
+      var topLeft = transform.transformPoint({x:0,y:0});
+      var bottomRight = transform.transformPoint({x:box.right, y:box.bottom});
+      box.top = topLeft.y;
+      box.left = topLeft.x;
+      box.right = bottomRight.x;
+      box.bottom = bottomRight.y;
```
davidaurelio, line 95: Re-calculation of `.width` and `.height` is needed here

src/runner/path/curved_path.js
```diff
@@ -570,6 +570,51 @@ define([
     return curvedSegments;
   };
 
+  /** 
+   * Calculates the potential bounds of a single cubic bezier curve
+   * @param {Array} p0 The starting point of the curve in the form [x, y]
+   * @param {Array} curve A curveTo segment (e.g. `['curveTo',n,n,n,n,n,n]`)
+   */
+  CurvedPath.getPotentialBoundsOfCurve = function(p0x, p0y, cp1x, cp1y, cp2x, cp2y, p1x, p1y) {
```
davidaurelio: What is potential about the returned bounds?

davidaurelio:
Awesome demo. I’ve added a few remarks. Can’t verify the bounds algorithm for cubic bezier curves (no time), but it seems to work.

davidaurelio:
ready for merge? further objections?

padolsey:
Just to note: This pull request does not fix https://github.com/uxebu/bonsai/issues/2

iamdustan:
@padolsey very nice demo, indeed! :thumbsup:
I think that would be worth adding to http://demos.bonsaijs.org/

PR #138, opened 2012-10-29
Implement `interactive` attr on DisplayObject

padolsey:
Using `pointer-events:...;` on the renderer-side to support this.

### DIFF
--- a/src/renderer/svg/svg.js
+++ b/src/renderer/svg/svg.js
@@ -139,6 +139,9 @@ define([
         }
 
         switch (i) {
+          case 'interactive':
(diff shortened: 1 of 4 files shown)


src/renderer/svg/svg.js
```diff
@@ -139,6 +139,9 @@ define([
         }
 
         switch (i) {
+          case 'interactive':
+            el.style.pointerEvents = value ? 'inherit' : 'none';
```
davidaurelio, line 143: would it be better to just set it to the empty string? I really don’t know, just asking

Issue #126, opened 2012-10-22
getBoundingBox of `new Path` vs `new Rect`


basecode:
Getting different results feels wrong:

```
var aPath = new Path('M20,20l150,0l0,150l-150,0').fill("red").addTo(stage);
console.log(aPath.getBoundingBox());  // left:20,top:20

var aRect = new Rect(20, 20, 150, 150).fill("green").addTo(stage);
console.log(aRect.getBoundingBox()); // left:0, top:0
```

These results imply that the user knows that we convert a `Rect` internally into `new Rect(0,0,150,150).attr('matrix', new Matrix(1,0,0,1, 20,20))`

Also, I think the bbox of `aRect` is needless. It actually just returns the size of the DisplayObject without any offset (without transformation). When I just want to know the size (width and height) of an object then I think we could offer sth like `DisplayObject.size()`. // { width, height}

padolsey:
Interesting topic.

I think our Rect implementation is fine. I think it would make no sense if this:

``` js
new Rect(20,20,100,100)
```

Did this:

``` js
M20,20l100,0l0,100l-100,0
```

I mean, when I call `new Rect` I want to draw a single self-contained rect. It makes no sense for me to want the rect to start at a local coordinate of `{20,20}`. 

I think that local coordinate space is mostly moot when you just want a single shape as a display object.

The question seems to be: _Why would a user want to manipulate the local starting coordinate (M20,20) of a single shape?_ -- I can't think of an answer

The x and y parameters for `new Rect` are, IMHO, far more useful for setting the position of the shape in its parent coordinate space.

---

There seems to be another thing we need to discuss, and it should probably be a separate discussion. `getBoundingBox` when called without arguments gives the bounding-box in the local coordinate space. This can seem to give strange results (as you've shown).

The most common `getBoundingBox` use-case will probably be to get the bbox of a path within its parent/global coordinate space. We currently offer no easy way to get the global bbox of a DisplayObject. I think this is a problem.

basecode:
Just to get the terminology right:

A "bounding box" is actually the offset and size of a DisplayObject in a certain coordinate system.
"Offset" is the position relative to the upper left corner of a parent context.

There are three interesting coordinate systems:
- Local (offset relative to the DisplayObject itself, e.g. triggered by a M20,20)
- Parent (offset relative to the parent container. e.g. a Group)
- Userspace / Global (offset relative to the bonsai-movie). Comparable with SVG's `getBBox` API 

[...]

PR #141, opened 2012-10-30
Implement getAbsoluteBoundingBox and getAbsoluteMatrix

padolsey:
Discussion point: Is it better to have a `getBoundingBoxRelativeTo`?

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,7 @@
 v0.4.2
 -------------------
-
+ * Added support for getAbsoluteBoundingBox and getAbsoluteMatrix on
+   DisplayObject
 
 v0.4.1 / 2012-10-26
 -------------------
--- a/src/runner/display_object.js
+++ b/src/runner/display_object.js
@@ -577,12 +577,37 @@ define([
       return this;
     },
 
+    /** 
+     * Gets the matrix of the DisplayObject combined with all ancestor matrices
+     *
+     * @returns {Matrix} The resulting matrix
+     */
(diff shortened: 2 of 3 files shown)


davidaurelio:
boundingBoxRelativeTo is: find common ancestor, compute bounding boxes relative to that, apply difference. Not sure whether we really need that

I think `getParentBoundingBox` / `getBoundingBoxParent` is really needed.

davidaurelio:
+1 for `getAbsoluteMatrix()`. That’s cool to have!

iamdustan:
Could I propose `getAbsoluteTransformOrigin` as well? :)

basecode:
Related: #126

basecode:
What I can see is that 
- our current implementation of `getBoundingBox` actually returns the local offset (#126 explains terminology)
- `getAbsoluteBoundingBox` returns the userspace resp. global offset. Exactly what SVG's `getBBox` is known for.
- `getParentBoundingBox` (proposed by @davidaurelio) would be neat to have.

I propose the following:
- rename `getBoundingBox` to `getLocalBBox(aMatrix)`
- rename `getAbsoluteBoundingBox` to `getBBox()` or `getAbsoluteBBox`
- rename `getParentBoundingBox` to `getParentBBox()`

`getBBox` is an equivalent to `getLocalBBox(this.getAbsoluteMatrix())`
`getParentBoundingBox` is equivalent to `getLocalBBox(this.attr('matrix'))`

PR #142, opened 2012-10-30
Calculate the angle between two points.

iamdustan:
Add a helper method to calculate the angle between two points.

### DIFF
--- a/src/point.js
+++ b/src/point.js
@@ -164,7 +164,7 @@ define(function() {
    * Returns the angle in radians or degrees between pt1 and pt2
    *
    * @param {Point} toPoint
-   * @param {asDegrees} [boolean=false]
+   * @param {Boolean} [asDegrees=false]
    * @returns {Number} angle
    */
   proto.angle = function(toPoint, asDegrees) {
--- a/src/point.js
+++ b/src/point.js
@@ -159,6 +159,23 @@ define(function() {
     return sqrt(hside * hside + vside * vside);
   };
 
+
+  /**
+   * Returns the angle in radians or degrees between pt1 and pt2
+   *
+   * @param {Point} toPoint
+   * @param {asDegrees} [boolean=false]
+   * @returns {Number} angle
+   */
+  proto.angle = function(toPoint, asDegrees) {
+    var radians = Math.atan2(this.y - toPoint.y, this.x - toPoint.x);
+    if (asDegrees) {
+      return radians * 180 / Math.PI;
+    } else {
+      return radians;
+    }
+  };
+
   // static methods
 
   /**
--- a/test/point-spec.js
+++ b/test/point-spec.js
@@ -103,6 +103,24 @@ define([
       });
     });
 
+    describe('angle', function() {
+
+      it('Returns the radians between two points', function() {
+        var a = new Point(0, 0),
+            b = new Point(180, 180),
+            radians = Math.atan2(0 - 180, 0 - 180);
+        expect(a.angle(b)).toBe(radians);
+      });
+
+      it('Returns the angle between two points', function() {
+        var a = new Point(0, 0),
+            b = new Point(-180, -180),
+            degrees = Math.atan2(0 - -180, 0 - -180)*180/Math.PI;
+        expect(a.angle(b, true)).toBe(degrees);
+      });
+
+    });
+
     // Linear Interpolation between two points a and b
     describe('lerp', function() {
       it('it is a class method', function() {


src/point.js
```diff
@@ -159,6 +159,23 @@ define(function() {
     return sqrt(hside * hside + vside * vside);
   };
 
+
+  /**
+   * Returns the angle in radians or degrees between pt1 and pt2
+   *
+   * @param {Point} toPoint
+   * @param {asDegrees} [boolean=false]
```
padolsey: I think this should be:

``` js
* @param {Boolean} [asDegrees=false]
```

padolsey:
Thanks for this addition Dustan! There's a tiny correction needed in the jsdoc comment and then it's fine to merge I think.

iamdustan:
Whoops! Updated.

davidaurelio:
Sorry, I’m too late.

I have a question and a remark:
1. What is the angle between two points? A third point is needed for an angle. Is it the angle through point _A, (0, 0),_ and _B_?
2. We try to avoid boolean arguments where possible. Even if we have some of them. The second parameter should be either `unit` (and return degrees if `'degrees'` is passed, or go away entirely, or be `throughPoint`, defaulting to _(0, 0)_

iamdustan:
1. The angle between two points is currently calculated with the positive X plane to the counterclockwise angle of `thisPoint` and `argumentPoint`. This may need to be beefed up to take a 2nd point to calculate the angle difference, but I would extend that basing it off of a cartesian coordinate understanding of angles would be appropriate. https://developer.mozilla.org/en-US/docs/JavaScript/Reference/Global_Objects/Math/atan2#Description
2. Would you like me to update 'degrees' or 'radians' to be passed as the second argument? Since radians is usually the default method for working with angles for non-humans the second argument could be changed out. The only compelling argument I have for `asDegrees` is to be able to get something my brain can understand back to verify that it is returning what I would expect.

The use case I have for this is to compute the rotation angle to apply between pointermove events in relation to an arbitrary Point.

To explain my `cartesian coordinate understanding of angles` statement (which is not clear at all)

Given a point difference of:
- +x, +y: the returned angle will be between +0 and +90,
- +x, -y: the returned angle will be between +90 and +180
- -x, -y: the returned angle will be between -90 and -180
- +x, -y: the returned angle will be between -0 and -90.

basecode:
We should consider to remove the second parameter and decide for either radians or degrees. I think that would make the Point class more readable, consistent and simple regarding the behavioral specification. The Point class already deals with signatures where only radians are allowed and not both (see `Point.polar`). 

Apart from that I would avoid duplicate code and prefer to use an external function (separately tested) for the process of converting radians to degrees and vice-versa. If not already available I would introduce `tools.radFromDeg` and `tools.degFromRad`.

I would be even more happy if we could agree on one default unit for the whole Bonsai API and not just the Point class. Compare `gradient.LinearGradient` that takes degrees and `Matrix#rotate()` that takes radians as parameter.

I would go with degrees b/c ppl are more familiar with degrees than with radians.

davidaurelio:
Now I get it. It calculates the **slope of a straight line.**

re second argument – I would prefer to leave it out altogether for now and rather add conversion functions to `tools`.

The problem that we really have is an evil mix of radians and radians throughout our API.

radians are used by:
- `DisplayObject` (rotation)
- `Arc` (startAngle, endAngle)

degrees are used by:
- `LinearGradient` (direction)
- `CurvedPath.fromArc` (angle) – also: unfinished oO

And I propably did not find everything. `Path` seems to be consistent by now (using radians everywhere)

iamdustan:
Degrees v Radians:
I agree degrees are much more intuitive to work with, but I am not a graphics programmer. It seems even in everyday html work a developer needs to go between radians and degrees regularly.

A few examples:

Radians:
- Canvas#rotating :: https://developer.mozilla.org/en-US/docs/Canvas_tutorial/Transformations#Rotating
- Canvas#arcs :: https://developer.mozilla.org/en-US/docs/Canvas_tutorial/Drawing_shapes#Arcs

Degrees
- CSS Linear Gradients: http://www.w3.org/TR/2011/WD-css3-images-20110712/#linear-gradient-syntax
- CSS Rotation Transforms: http://davidwalsh.name/css-transform-rotate

What I gather from this is that lower level graphics developers are expected or assumed to be comfortable enough with radians to work (or write their own methods around it), where as CSS developers are given the benefit of using the friendlier degrees.

Not to promote bastardizing the Bonsai API, but what are the thoughts of taking a lead from the above examples and accepting both in all places via string arguments such as `'37deg'` or radians `0.645771823`. I feel as this wouldn’t be a far cry off of the animation class acception arguments in ms, Ns, or N%.

The conversion methods would be quite straightforward.

```
var toDegrees = 180 / Math.PI,
    toRadians = Math.PI / 180;

tools.radFromDeg = function(deg) {
  return deg * toRadians;
}

tool.degFromRad = function(radians) {
  return radians * toDegrees;
}
```

iamdustan:
Hmm....I’m already detecting a potential issue of what that proposal could cause internally. Would the helpers return `Number`’s or something similar to the expected argument style.

tools.radFromDeg = function(deg) {
  return parseFloat(deg) \* toRadians;
}
tools.degFromRad = function(radians) {
  return (radians \* toDegrees) + 'deg');
}

basecode:
> Not to promote bastardizing the Bonsai API, but what are the thoughts of taking a lead from the above examples and accepting both in all places via string arguments such as '37deg' or radians 0.645771823. I feel as this wouldn’t be a far cry off of the animation class acception arguments in ms, Ns, or N%.

Like it! CSS supports '37deg' and '0.645771823rad' but I also think it makes more sense to treat rad as Number.
I also think it makes more sense when `degFromRad` returns a string. Well spotted!

Nevertheless, I don't think this rule applies to the Point class. At least for me the Point class should be seen as a separated, isolated and also more low-level library.

iamdustan:
Ah. I didn’t even realize you could use other unit types in css.
http://www.w3.org/TR/css3-values/#angles

Proposed plan of action would:

All angle based arguments accept a string 'ndeg' for degrees or a Number for radians and is responsible for converting to the correct type it needs for any lower level calls.

Therefore, those two helpers should be created to return `degFromRad` and `radFromDeg` to begin work on normalizing degree/radian arguments throughout the API to accept both.

PR #144, opened 2012-11-02
Make sure touchstart+touchend do only trigger ONE click event, not two.

wolframkriesing:
See the example, before it used to fire two clicks.

### DIFF
--- a/example/library/movies/event-click.js
+++ b/example/library/movies/event-click.js
@@ -1,9 +1,7 @@
 /**
- * event click
- *
  * Path w/ events.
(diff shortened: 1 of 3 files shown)


wolframkriesing:
so this breaks drag and probably multitouch ... help appreciated, like pointers in the right direction or alike, thx
i guess some tests that show that drag+multitouch breaks would be best

davidaurelio:
Superseded by 4ac85d2b805e107908c866250179f69b66710a2e

PR #148, opened 2012-11-09
Just trigger on click for touchstart+end

wolframkriesing:
this branch contains the missing test for the fix submited in https://github.com/uxebu/bonsai/commit/4ac85d2b805e107908c866250179f69b66710a2e and examples that show the fixed behavior


davidaurelio:
I guess from reading the code, that the supplied test case might not fail with the old code. I will take the time to change it later.

wolframkriesing:
I tried  it with the old code and it failed
Am 09.11.2012 11:15 schrieb "David Aurelio" notifications@github.com:

wolframkriesing:
But please doubke check

davidaurelio:
Good that I wrote “guess” and “might” ;-) just read through it, will try it out later.

Issue #134, opened 2012-10-26
Implement stroke-dasharray


iamdustan:
Add `strokeDashArray` to path supported attributes: http://docs.bonsaijs.org/module-path.Path.html

**Documentation**
https://developer.mozilla.org/en-US/docs/SVG/Attribute/stroke-dasharray

**Example**
https://developer.mozilla.org/files/3299/stroke-dasharray.svg

basecode:
Implemented in version 0.4.3.

PR #153, opened 2012-11-16
stroke-dasharray attribute on paths

iamdustan:
I’ve started work on #134 to implement the stroke-dasharray attribute.
(docs: https://developer.mozilla.org/en-US/docs/SVG/Attribute/stroke-dasharray)

I believe everything should be set on the runner side of things, but I’m looking for some pointers on where to go in the `renderer/svg/` files to implement the receiving side of things and what the easiest way to visually and jasmine-y test it.

### DIFF
--- a/src/renderer/svg/svg.js
+++ b/src/renderer/svg/svg.js
@@ -47,6 +47,7 @@ define([
     opacity: 'opacity',
     fillOpacity: 'fill-opacity',
     strokeOpacity: 'stroke-opacity',
+    strokeDashArray: 'stroke-dasharray',
     fontSize: 'font-size',
     fontWeight: 'font-weight',
     fontStyle: 'font-style',
--- a/src/renderer/svg/svg.js
+++ b/src/renderer/svg/svg.js
@@ -65,6 +65,7 @@ define([
     fillOpacity: ['fill-opacity', '1'],
     strokeOpacity: ['stroke-opacity', '1'],
     strokeDashArray: ['stroke-dasharray'],
+    strokeOffset: ['stroke-dashoffset'],
     fontSize: ['font-size'],
     fontWeight: ['font-weight'],
     fontStyle: ['font-style'],
@@ -562,6 +563,11 @@ define([
       if ('strokeGradient' in attr) {
         this.applyStrokeGradient(element, attr.strokeGradient, '', attr.strokeWidth);
       }
+
+      if ('strokeDashArray' in attr) {
+        this.applyStrokeDashArray(element, attr.strokeDashArray, attr.strokeOffset || 0);
+      }
+
     }
   };
 
@@ -1018,6 +1024,16 @@ define([
     } else if (aColor === null) {
       element.removeAttribute('stroke');
       element.removeAttribute('data-stroke');
+    }
+  };
+
+  proto.applyStrokeDashArray = function(element, strokeDashArray, strokeOffset) {
+    if (strokeDashArray) {
+      element.setAttribute('stroke-dasharray', strokeDashArray);
+      element.setAttribute('stroke-dashoffset', strokeOffset);
+    } else {
+      element.removeAttribute('stroke-dasharray');
+      element.removeAttribute('stroke-dashoffset');
     }
   };
 
--- a/src/renderer/svg/svg.js
+++ b/src/renderer/svg/svg.js
@@ -565,7 +565,7 @@ define([
       }
 
       if ('strokeDash' in attr) {
-        this.applyStrokeDashArray(element, attr.strokeDash.join(''), attr.strokeDashOffset);
+        this.applyStrokeDashArray(element, attr.strokeDash, attr.strokeDashOffset);
       }
 
     }
--- a/src/renderer/svg/svg.js
+++ b/src/renderer/svg/svg.js
@@ -64,8 +64,8 @@ define([
     opacity: ['opacity', '1'],
     fillOpacity: ['fill-opacity', '1'],
     strokeOpacity: ['stroke-opacity', '1'],
-    strokeDashArray: ['stroke-dasharray'],
-    strokeOffset: ['stroke-dashoffset'],
+    strokeDash: ['stroke-dasharray'],
+    strokeDashOffset: ['stroke-dashoffset'],
     fontSize: ['font-size'],
     fontWeight: ['font-weight'],
     fontStyle: ['font-style'],
@@ -564,8 +564,8 @@ define([
         this.applyStrokeGradient(element, attr.strokeGradient, '', attr.strokeWidth);
       }
 
-      if ('strokeDashArray' in attr) {
-        this.applyStrokeDashArray(element, attr.strokeDashArray, attr.strokeOffset || 0);
+      if ('strokeDash' in attr) {
+        this.applyStrokeDashArray(element, attr.strokeDash.join(''), attr.strokeDashOffset);
       }
 
     }
--- a/src/renderer/svg/svg.js
+++ b/src/renderer/svg/svg.js
@@ -1030,7 +1030,7 @@ define([
   proto.applyStrokeDashArray = function(element, strokeDashArray, strokeOffset) {
     if (strokeDashArray) {
       element.setAttribute('stroke-dasharray', strokeDashArray);
-      element.setAttribute('stroke-dashoffset', strokeOffset);
+      element.setAttribute('stroke-dashoffset', strokeOffset || 0);
     } else {
       element.removeAttribute('stroke-dasharray');
       element.removeAttribute('stroke-dashoffset');
--- a/src/runner/animation/translators/stroke_dash.js
+++ b/src/runner/animation/translators/stroke_dash.js
@@ -28,9 +28,6 @@ define(function() {
           if (isNaN(strokeDash)) {
             strokeDashes[i] = 0;
           }
-          else {
-            strokeDashes[i] = strokeDash;
-          }
         }
 
         return strokeDashes;
--- a/src/runner/animation/translators/stroke_dash.js
+++ b/src/runner/animation/translators/stroke_dash.js
@@ -1,0 +1,41 @@
+define(function() {
+  'use strict';
+
+  /**
+   * Animation translations for shape segments
+   * @ignore
+   */
+
+  return {
+    strokeDash: {
+      toNumeric: function(strokeDashes, isEndValues) {
+        var numbers = [];
+        if (!isEndValues) {
+          this.strokeDash = strokeDashes;
+        }
+        for (var i = 0, l = strokeDashes.length; i < l; ++i) {
+          var strokeDash = strokeDashes[i];
+          if (!isNaN(strokeDash)) {
+            numbers.push(strokeDash);
+          }
+        }
+        return numbers;
+      },
+      toAttr: function(numbers) {
+        var strokeDashes = numbers;
+        for (var i = 0, l = strokeDashes.length; i < l; i++) {
+          var strokeDash = strokeDashes[i];
+          if (isNaN(strokeDash)) {
+            strokeDashes[i] = 0;
+          }
+          else {
+            strokeDashes[i] = strokeDash;
+          }
+        }
+
+        return strokeDashes;
+      }
+    }
+  };
+});
+
--- a/src/runner/animation/translators/stroke_dash.js
+++ b/src/runner/animation/translators/stroke_dash.js
@@ -9,23 +9,18 @@ define(function() {
   return {
     strokeDash: {
       toNumeric: function(strokeDashes, isEndValues) {
-        var numbers = [];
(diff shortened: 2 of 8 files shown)


iamdustan:
To copy some things into this from a conversation with @basecode on IRC:

### dasharray offset

Spec: http://www.w3.org/TR/SVG/painting.html#StrokeDashoffsetProperty

### Canvas implementation
- State today: https://developer.mozilla.org/en-US/docs/DOM/CanvasRenderingContext2D#Gecko-specific_attributes
- Spec: http://www.whatwg.org/specs/web-apps/current-work/#line-styles

halida:
stroke-dasharray is a **must** feature, hope it will be accepted soon.

iamdustan:
@halida I started work on this a while back. I’ll pick it back up in the next couple weeks. I’ll try to remember to ping you whenever I start to work on it again.

halida:
@iamdustan Cool!

iamdustan:
Well look at that. Two months later I finish it. There are tests for the runner.

I was going to test the renderer side of things, but it seemed that the test file for that is in a bit of disarray:
https://github.com/uxebu/bonsai/blob/master/test/renderer/svg_renderer_svg-spec.js#L59-L105

Instead I just added the example movie with 550522c

src/runner/path/path.js
```diff
@@ -210,6 +232,11 @@ define([
       strokeColor: accessor(getStrokeColor, setStrokeColor, true),
       _strokeGradient: data(undefined, true),
       strokeGradient: accessor(getStrokeGradient, setStrokeGradient, true),
+      _strokeDashArray: data(null, true),
+      strokeDashArray: accessor(getStrokeDashArray, setStrokeDashArray, true),
+      _strokeOffset: data(null, true),
```
basecode: It should be fine handling `strokeOffset` similar to `strokeWidth` and making use of the default getter/setter.
`strokeOffset: data(0, true, true)`

src/runner/path/path.js
```diff
@@ -101,6 +101,27 @@ define([
     this._strokeColor = parseColor(color, this._strokeColor);
   }
 
+  var getStrokeDashArray = getter('_strokeDashArray');
+  function setStrokeDashArray(dashArray) {
+    if (dashArray) {
+      if (dashArray.join) {
```
basecode: Please use `tools.isArray` instead. For "keeping things simple", I would suggest to only allow "array" or "null / empty array" for now. Do you think supporting strings is a "must" feature?

src/runner/path/path.js
```diff
@@ -180,6 +201,7 @@ define([
    * @property {String} __supportedAttributes__.join The shape to be used at the corners of paths. Can be one of 'miter', 'round', 'bevel'. Default: 'miter'
    * @property {String} __supportedAttributes__.strokeColor The line color. Default: transparent
    * @property {gradient.LinearGradient|gradient.RadialGradient} __supportedAttributes__.strokeGradient The line gradient. Default: nothing
+   * @property {String} __supportedAttributes__.strokeDashArray The stroke dasharray. Default: nothing
```
basecode: - I would go with allowing `{Array|null}` instead of `{String}`. 
- Even so I like explicit naming, I consider `strokeDashArray` as too long. I would prefer `strokeDash`. It also conflicts with the Type "Array". It's quite possibly that we support strings as well at some point.
- I would change `Default:nothing` to `Default:null`. Not sure whether we're using the term "nothing" somewhere else?
- Please add a line that describes `strokeOffset`.

test/path/path-spec.js
```diff
@@ -323,6 +323,15 @@ define([
         expect(s.attr('strokeColor')).toBe(0xff0000ff);
       });
 
+      it('Sets & Gets strokeDashArray', function(){
```
basecode: Unfortunately we started to put all expectations in one "it" block for some reasons. Please try not to copy that pattern. :)

Instead write:

```
describe('Sets & Gets strokeDashArray', function() {
  it('returns null by default', function() {
    expect(new Path().attr('strokeDashArray')).toBeNull();
  });
  …
});
```

basecode: Could you please add tests for `strokeOffset`?

basecode: Good example: https://github.com/uxebu/bonsai/blob/master/test/display_object-spec.js#L233

src/runner/path/path.js
basecode: In case "offset" only affects "strokeDash" and not "stroke" itself, I would propose to rename "strokeOffset" to "strokeDashOffset".

basecode:
Thanks, @iamdustan! As you can see, I started reviewing it :) Since this change introduces a new morphable property, it would suggest to also add a morphing example. Maybe in the same file. What do you think?

src/runner/path/path.js
iamdustan: No, I don’t think it’s a must. Looking at the SVG spec it’s represented as a string, and the canvas implementation seems to be only represented as an Array.

src/runner/path/path.js
iamdustan: - Switch to {Aray|null} :thumbsup:
- `strokeDash` instead of `strokeDashArray` :thumbsup: (ironically, the canvas spec which works with an array doesn’t have array in the name, and the svg spec which uses strings has `array` in the property name)
- I took `Default:nothing` from the line right above that.

```
* @property {gradient.LinearGradient|gradient.RadialGradient} __supportedAttributes__.strokeGradient The line gradient. Default: nothing
```
- `strokeOffset` description :thumbsup:

src/runner/path/path.js
iamdustan: I think that the `strokeOffset` could be applied with or without a dash, but there would be no visual impact. I’m game for the explicitness of `strokeDashOffset`.

Also, thanks for the tip on the default getters/setters; was not aware of that.

iamdustan:
Thanks for the review @basecode! I appreciate your bravery to enter into my pull requests. They always end up being a marathon run ;)

I assume “morphable” means capable of being animated? I will add and I see no reason not to put it in the same file.

iamdustan:
@basecode code changes are up

iamdustan:
@basecode I updated the demo movie with an animation of the strokeDashOffset, but the strokeDash wasn’t getting animated appropriately. `undefined` was getting passed through.

(moments later). Ah. I take `src/runner/animation/translators/stroke-dash.js` needs to be created?

test/path/path-spec.js
```diff
@@ -323,6 +323,32 @@ define([
         expect(s.attr('strokeColor')).toBe(0xff0000ff);
       });
 
+      describe('Sets & Gets strokeDash', function(){
+        it('returns null be default', function(){
+          expect(new Path().attr('strokeDash')).toBe(null);
+        });
+
+        it('should be settable/gettable', function(){
```
basecode, line 331: Please turn this into two tests (one that tests the getter and one the setter).

src/renderer/svg/svg.js
```diff
@@ -561,6 +563,11 @@ define([
       if ('strokeGradient' in attr) {
         this.applyStrokeGradient(element, attr.strokeGradient, '', attr.strokeWidth);
       }
+
+      if ('strokeDash' in attr) {
+        this.applyStrokeDashArray(element, attr.strokeDash.join(''), attr.strokeDashOffset);
```
basecode: The "in" operator returns true even when attr.strokeDash is `null`. Since `attr.strokeDash` is null, `join` would throw. I think you can safely remove the "join" call. In L1032 (`element.setAttribute('stroke-dasharray', strokeDashArray);`) the array should get casted to a string anyway.

basecode:
@iamdustan re:animation. Yes, since strokeDash isn't numeric, it can't be "auto translated". After creating `src/runner/animation/translators/stroke-dash.js` you also need to announce it [here](https://github.com/uxebu/bonsai/blob/master/src/runner/animation/properties_tween.js#L31).

test/path/path-spec.js
iamdustan, line 331: Doesn’t `return null by default` test the getter implicitly?
And the only way to test the setter seems to be by using the getter, no, and the only way to test the getter (beyond the default) is by using the setter...

src/renderer/svg/svg.js
iamdustan: Good call. Who would’ve thought browsers were smart enough to `toString` the applied attribute?

test/path/path-spec.js
basecode, line 331: I guess my comment was a little bit confusing. What I was trying to say is that an "it" block should only contain one expectation (L333, L335).

iamdustan, line 331: Ah. So don’t copy the other example exactly ;) I’ll remove the second one. The first `expect` should fully test both getter and setter.

iamdustan:
@basecode see f4097ed for my first stab at a translator. There is an example included in everything looks to be working correctly.

src/runner/animation/translators/stroke_dash.js
```diff
@@ -0,0 +1,38 @@
...
+        for (var i = 0, l = strokeDashes.length; i < l; ++i) {
+          var strokeDash = strokeDashes[i];
+          if (!isNaN(strokeDash)) {
+            numbers.push(strokeDash);
+          }
+        }
+        return numbers;
+      },
+      toAttr: function(numbers) {
+        var strokeDashes = numbers;
```
basecode: Why this extra assignment?

src/runner/animation/translators/stroke_dash.js
```diff
@@ -0,0 +1,38 @@
...
+
+  return {
+    strokeDash: {
+      toNumeric: function(strokeDashes, isEndValues) {
+        var numbers = [];
+        if (!isEndValues) {
+          this.strokeDash = strokeDashes;
+        }
+        for (var i = 0, l = strokeDashes.length; i < l; ++i) {
+          var strokeDash = strokeDashes[i];
```
basecode: Please try to only declare variables in scoped blocks. The body of the for-loop isn't scoped.  Put `var strokeDash;` either in L12 or L16.

src/runner/animation/translators/stroke_dash.js
```diff
@@ -0,0 +1,38 @@
...
+   * Animation translations for shape segments
+   * @ignore
+   */
+
+  return {
+    strokeDash: {
+      toNumeric: function(strokeDashes, isEndValues) {
+        var numbers = [];
+        if (!isEndValues) {
+          this.strokeDash = strokeDashes;
```
basecode: Wondering why we need this at all. Since `this.strokeDash` isn't used in `toAttr`. I think it's okay to remove L13-15. See: https://github.com/uxebu/bonsai/blob/master/src/runner/animation/property_tween.js#L28

src/runner/animation/translators/stroke_dash.js
iamdustan: Very good question. I copied it from https://github.com/uxebu/bonsai/blob/master/src/runner/animation/translators/segment.js#L28 without fully understanding why, but I also didn’t understand what it was doing at the time.

src/runner/animation/translators/stroke_dash.js
iamdustan: I have often found myself doing things in bonsai commits I wouldn’t do in my normal development, but I tried to stick to the coding style of the files around it. (See https://github.com/uxebu/bonsai/blob/master/src/runner/animation/translators/segment.js again).

Though now that I’ve gone through the other translators, it seems that the segment translator is the only one that includes variable declaration within for loops.

src/runner/animation/translators/stroke_dash.js
iamdustan: Ah! Very nice. That definitely sheds some light on how the `toNumeric` and `toAttr` are connected together.

iamdustan:
@basecode I greatly decreased the complexity of the translator in 4e5a66e650263feafc28b781912a013b91ba027f to simply push all real numbers to the array in `toNumeric` and reset NaN’s to 0 in the `toAttr` call.

basecode:
Looks fine to me. Thanks for that great contribution @iamdustan! One last thing. Please add a line to the CHANGELOG and we're done.

basecode:
Additional info: The current implementation implicitly leaks SVG specific API. 

"If an odd number of values is provided, then the list of values is repeated to yield an even number of values. 
Thus, 5,3,2 is equivalent to 5,3,2,5,3,2."

But since all other possible drawing APIs are probably following this spec as well I don't see a reason for solving that problem on the runner side right now.

[Canvas API](http://www.whatwg.org/specs/web-apps/current-work/#dom-context-2d-setlinedash) says: "If the number of elements in a is odd, then let a be the concatentation of two copies of a."

iamdustan:
Changelog updated.

iamdustan:
:thumbsup: We made it!

PR #157, opened 2012-11-29
Renderer frame times

davidaurelio:
Prevents a memory leak (frame times) and increasing runtime (pushing onto an ever growing array)

### DIFF
--- a/src/renderer/svg/svg.js
+++ b/src/renderer/svg/svg.js
@@ -237,6 +237,7 @@ define([
     document.addEventListener('keydown', this, false);
     document.addEventListener('keypress', this, false);
 
+    this._isLoggingFps = false;
     this._setupFPSLog(options.fpsLog);
     if (options.disableContextMenu) {
       this.config({
--- a/src/renderer/svg/svg.js
+++ b/src/renderer/svg/svg.js
@@ -422,7 +422,9 @@ define([
       }
     }
 
-    this._logFrame();
+    if (this._isLoggingFps) {
+      this._logFrame();
+    }
 
     this.emit('canRender');
   };
@@ -1490,7 +1492,10 @@ define([
 
   proto._setupFPSLog = function(fpsLog) {
     var isFunction = typeof fpsLog === 'function';
-    if (fpsLog !== true && !isFunction) {
+    var hasLog = fpsLog === true && !isFunction;
+    this._isLoggingFps = hasLog;
+    clearInterval(this._fpsInterval);
+    if (hasLog) {
       return;
     }
 
--- a/src/renderer/svg/svg.js
+++ b/src/renderer/svg/svg.js
@@ -1490,7 +1490,7 @@ define([
 
   proto._setupFPSLog = function(fpsLog) {
     var isFunction = typeof fpsLog === 'function';
-    if (fpsLog !== true || isFunction) {
+    if (fpsLog !== true && !isFunction) {
       return;
     }
 
--- a/src/renderer/svg/svg.js
+++ b/src/renderer/svg/svg.js
@@ -1498,14 +1498,14 @@ define([
 
   proto._setupFPSLog = function(fpsLog) {
     var isFunction = typeof fpsLog === 'function';
-    var hasLog = fpsLog === true && !isFunction;
+    var hasLog = fpsLog === true || isFunction;
     this._isLoggingFps = hasLog;
     clearInterval(this._fpsInterval);
-    if (hasLog) {
+    if (!hasLog) {
       return;
     }
 
-    if (!isFunction) {
+    if (fpsLog === true) { // draw a fps counter on the stage
       this.render([
         {
           parent:0,
--- a/src/renderer/svg/svg.js
+++ b/src/renderer/svg/svg.js
@@ -827,6 +827,8 @@ define([
 
   };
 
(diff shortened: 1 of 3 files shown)


src/renderer/svg/svg.js
```diff
@@ -1493,7 +1497,10 @@ define([
 
   proto._setupFPSLog = function(fpsLog) {
     var isFunction = typeof fpsLog === 'function';
-    if (fpsLog !== true || isFunction) {
+    var hasLog = fpsLog === true && !isFunction;
+    this._isLoggingFps = hasLog;
+    clearInterval(this._fpsInterval);
+    if (hasLog) {
```
basecode: Since we return early here, can't  we remove the `if (!isFunction) {`?

src/renderer/svg/svg.js
```diff
@@ -1493,7 +1497,10 @@ define([
 
   proto._setupFPSLog = function(fpsLog) {
     var isFunction = typeof fpsLog === 'function';
-    if (fpsLog !== true || isFunction) {
+    var hasLog = fpsLog === true && !isFunction;
```
basecode: Much cleaner. +1

src/renderer/svg/svg.js
```diff
@@ -1493,7 +1497,10 @@ define([
...
+    this._isLoggingFps = hasLog;
```
basecode, line 1503: This adds silently a new instance variable at runtime. (Same with `this._frameTimes`) I see three reasons why we should avoid that and instead introduce all instance-vars in the constructor or via "proto".  

1) Readability. The Reader can't see all instance-vars at a glance and also can't see the expected default value.
2) We should not encourage ourself to "pseudo-hide" code b/c it is not important or just an opt-in function. 
3) Conflicts with existing strategies of JS-Engines to optimise the execution of JS.

davidaurelio, line 1503: 1) Agreed, but the SVG renderer has code like this all over the place. It mixes in stuff from other modules to be functional.
2) what do you mean by “pseudo-hide”
3) I’m not even sure runtimes create a hidden class for a type that has (in most cases) only one instance. Do you know more?

test/renderer/svg-spec.js
```diff
@@ -145,5 +145,45 @@ define([
...
+      }
+
+      function createClock(fps) {
+        var lastTime = 0, interval = 1000 / fps;
+        return function() {
+          return Math.round(lastTime += interval);
+        }
+      }
+
+      it('should not collect frame times that are older than one second, even if fps logging is disabled', function() {
```
basecode: s/disabled/enabled ? I would expect frame times are not collected when fps logging is disabled.

src/renderer/svg/svg.js
davidaurelio, line 1503: I’ve added a commit that initializes the property in the constructor

test/renderer/svg-spec.js
```diff
@@ -145,5 +145,45 @@ define([
...
     });
 
+    describe('Frame logging', function() {
+      function createSvgRenderer(fpsLog, getTime) {
+        var renderer = new SvgRenderer(createFakeDomNode(), 1, 1, {fpsLog: fpsLog});
+        renderer.getTime = getTime;
+        return renderer;
+      }
+
+      function createClock(fps) {
```
basecode, line 155: +1

src/renderer/svg/svg.js
davidaurelio: No, since we still need to distinguish between displaying an fps counter (`fpsLog === true`) and calling a callback with the framerate each second (`typeof fpsLog === 'function'`)

test/renderer/svg-spec.js
davidaurelio: Yes, but we did. And never threw them away. That’s the motivation for the test.

src/renderer/svg/svg.js
basecode, line 1503: 1) I know, but that doesn't mean that we should go on with that approach?
2) By "pseudo-hide" I mean code, that is not relevant for the normal use-case. (Couting FPS) Sth that could be moved in an isolated module in most cases. 
3) Afaik, no. In case of "hidden classes" a new class is created (that inherits from an existing hidden class) when ever an unknown property is added. In this case that would mean the "old" class w/o the "_isLoggingFps" is the new parent of the a newly created hidden class _with_ the new inst-var "_isLoggingFps". From now on the new class is used. This is really just relevant for V8's hidden classes. https://developers.google.com/v8/design#prop_access

davidaurelio, line 1503: 1.) You’re right :) The commit is already there
2.) Hm, fps logging is kind of baked in atm, so it might be difficult to move it out

src/renderer/svg/svg.js
```diff
@@ -1540,14 +1548,14 @@ define([
   };
 
   proto._logFrame = function() {
-    (this._frameTimes || (this._frameTimes = [])).push(+new Date);
+    (this._frameTimes || (this._frameTimes = [])).push(this.getTime());
```
basecode, line 1552: You could move `this._frameTimes` in the constructor `this._frameTimes = [];` and you could get rid of the existence-check and assignment here.

Issue #100, opened 2012-10-04
Reliable way to listen to a runner `ready` event


iamdustan:
There is no reliable way to detect when a runner environment and all plugins are loaded and ready to go. The `bonsai.run().on(`start`)` event behaves inconsistently if there are plugins to load.

I have created two demos to show the issue. Both demos do the exact same thing with the only change to the second demo is loading in a plugin.

**Demo 1. Expected behavior.**
http://jsfiddle.net/LAuFB/10/

**Demo 2. Message sent at `stage.on('start')` never received**
http://jsfiddle.net/LAuFB/9/

I started digging into the code and got in so far as  these two bits: https://github.com/uxebu/bonsai/blob/master/src/renderer/renderer_controller.js#L206
https://github.com/uxebu/bonsai/blob/master/src/renderer/renderer_controller.js#L60

A few solutions I have considered are:
1. Don’t fire the `start` event until all plugins are loaded.
2. Add a `ready` event once all plugins are loaded.
3. do a `stage.once('render', function () { stage.sendMessage('ready'); })` in userland (don’t fix)

davidaurelio:
Thanks for reporting, we will fix that with solution 1, I guess

iamdustan:
Fixed by #158. Closing.

Issue #150, opened 2012-11-14
on('load') doesn't work as expected on iOS


mattetti:
Take a look at this reduction: http://jsfiddle.net/XQdLu/

```
<!doctype html>
<html lang="en" class=" -webkit-">
  <head>
    <meta charset="utf-8" />
    <meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1, user-scalable=0;">
    <title>Demo</title>
    <script src="http://cdnjs.cloudflare.com/ajax/libs/bonsai/0.4.1/bonsai.min.js"></script>
  </head>

  <body>
    <div id="wrapper"></div>
  </body>

  <script>
  var movie = bonsai.run(document.getElementById('wrapper'), {
    code: function() {
      stage.on('message', function(data) {
        new Rect(10, 10, 100, 100)
        .fill('yellow')
        .stroke('#f00', 2)
        .addTo(stage);
      });
    },
    width: 400,
    height: 400,
  });

  movie.on('load', function() {
    this.sendMessage({ foo: 'bar' });
  });
  </script>
</html>
```

In my browsers (Chrome/Safari), the 'load' event gets triggered which sends the message to the movie variable which then draws the yellow Rect instance. However on my cellphone (iPhone running iOS5/safari, the rectangle never gets drawn.

iamdustan:
Potentially related: #100.

mattetti:
Note that this issue only seems to affect iOS 5, the same code works on iOS6

davidaurelio:
Ok, #158 does not fix this.

The issue is bound to the iframe runner, which is used by iOS5. I will debug that now and add a commit to #158

davidaurelio:
Found the cause: There is an expectation that code passed to `run()` can be executed synchronously. If that’s the case – as with the webworker runner – everything is fine. The iframe runner is not able to execute code synchronously, and the `'load'` event fires too early.

This issue will be adressed in the upcoming rewrite of the script loading system.

PR #158, opened 2012-11-29
Defer sendmessage

### LINKED ISSUE #154: cross movie messages sometimes fail
Here is a simple reduction of the bug:

http://jsfiddle.net/vnFLH/1/

Source code:
https://gist.github.com/4123564

Note that the `on load` callback function is always triggered but the message doesn't always get sent (which is the bug).

Tested in Chrome 23.0.1271.64 and Safari 6.0.2 (8536.26.17) on OS X 10.8.2

davidaurelio:
This defers sending of user messages to the runner context through `sendMessage()` until the runner has started up and is able to receive them.

Would fix #154

### DIFF
--- a/src/renderer/renderer_controller.js
+++ b/src/renderer/renderer_controller.js
@@ -11,6 +11,22 @@ define([
 function(tools, EventEmitter, URI) {
   'use strict';
 
+  function onContextLoad(rendererController) {
+    var pendingMessages = rendererController._pendingMessages;
+
+    // set to null first, so that further calls to sendMessage are not queued
+    rendererController._pendingMessages = null;
+
+    // send all queued messages again
+    var sendMessage = rendererController.sendMessage;
+    tools.forEach(pendingMessages, function(messageArguments) {
+      sendMessage.apply(rendererController, messageArguments);
+    });
+
+    // emit load event
+    rendererController.emit('load');
+  }
+
   var hitch = tools.hitch;
 
   /**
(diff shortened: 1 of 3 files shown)


iamdustan:
I am pretty sure this also fixes my issue in #100 and potentially #150?

davidaurelio:
If the `'load'` event is acceptable, this fixes #100, will check #150 when it is tomorrow at my place :)

iamdustan:
I’ll pull this branch in tomorrow and remove the code I’ve added to work around it. If it does the docs https://github.com/uxebu/bonsai-docs/blob/gh-pages/overview/Communication.md here (I think) will be able to be updated. I’ll verify tomorrow.

iamdustan:
:thumbsup: Just tested it and it is working like a charm.

src/renderer/renderer_controller.js
```diff
@@ -310,6 +327,14 @@ function(tools, EventEmitter, URI) {
      * @returns {this} The instance
      */
     sendMessage: function(category, messageData) {
+      var pendingMessages = this._pendingMessages;
+      if (pendingMessages) {
```
klipstein, line 341: could you be more explicit here? it probably is sufficient to add a comment which describes that it is set to null when the runner is ready.

davidaurelio, line 341: of course

mattetti:
Very nice, thank you very much!

PR #163, opened 2012-12-02
scaleX/scaleY updated when matrix is manipulated directly

### LINKED ISSUE #162: getScaleX/getScaleY return 1 if scale is applied through a matrix
First issue, assigning a scale via the matrix attribute, then getting the scale doesn’t return the applied scale.

```
myDisplayObject.attr('matrix', new Matrix(0.5, 0, 0, 0.5, 0 0));
`myDisplayObject.attr('scaleX')` // returns 1. Expects 0.5
```

Additionally

```
myDisplayObject.attr('scaleX', 0.5); // if applying a scale without using the matrix setters causes all future calls to setMatrix do nothing
myDisplayObject.attr('matrix', new Matrix(1, 0, 0, 1, 0, 0));
```

Is this expected?

iamdustan:
Fixes #162

I was hoping to simplify this down to be similar to the x/y getters and setters (manipulate the matrix directly), but my first attempt caused a large number of failing specs, so I opted for the double-setting in the `setMatrix` call instead.

I added some tests (didn’t push them) to ensure the same behavior for rotation, but they already passed. I could commit them to prevent potential regressions, but seeing as the rotation logic lives within the matrix class I didn’t think that was very likely.

### DIFF
--- a/src/runner/display_object.js
+++ b/src/runner/display_object.js
@@ -88,12 +88,16 @@ define([
 
   function setMatrix(matrix) {
     var m = this._matrix;
-    m.a = matrix.a;
+    m.a = this._scaleX = matrix.a;
     m.b = matrix.b;
     m.c = matrix.c;
-    m.d = matrix.d;
+    m.d = this._scaleY = matrix.d;
     m.tx = matrix.tx;
     m.ty = matrix.ty;
+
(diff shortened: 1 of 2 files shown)


davidaurelio:
Thank you for the fix! I’ve dded explanations for you to #162

PR #165, opened 2012-12-03
Rad/degree conversion helpers

iamdustan:
As per discussion at: https://github.com/uxebu/bonsai/pull/142#issuecomment-9942288

### DIFF
--- a/src/tools.js
+++ b/src/tools.js
@@ -345,10 +345,16 @@ define([], function() {
      * Get degrees value from a radian.
      *
      * @param {Number} rad radians
+     * @param {Number} [precision] Floating point precision
      * @returns {String} degrees
      */
-    degFromRad: function (rad) {
-      return (rad * Math.PI / 180) + 'deg';
+    degFromRad: function (rad, precision) {
+      var degrees = rad * 180 / Math.PI;
+      if (precision) {
+        degrees.toFixed(precision);
+      }
+
+      return degrees + 'deg';
     },
 
     /**
@@ -358,8 +364,9 @@ define([], function() {
      * @returns {Number} radians
      */
     radFromDeg: function (deg) {
-      return parseFloat(deg) * 180 / Math.PI;
+      return parseFloat(deg) * Math.PI / 180;
     }
+
   };
 
   return tools;
--- a/src/tools.js
+++ b/src/tools.js
@@ -339,6 +339,26 @@ define([], function() {
       array.length -= numRemoved;
 
       return array;
+    },
+
+    /**
+     * Get degrees value from a radian.
+     *
+     * @param {Number} rad radians
+     * @returns {String} degrees
+     */
+    degFromRad: function (rad) {
+      return (rad * Math.PI / 180) + 'deg';
+    },
+
+    /**
+     * Calculates the radians of a degree string
+     *
+     * @param {String|Number} deg degrees
+     * @returns {Number} radians
+     */
+    radFromDeg: function (deg) {
+      return parseFloat(deg) * 180 / Math.PI;
     }
   };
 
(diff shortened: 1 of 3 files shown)


basecode:
Great, thanks! Could you add some tests and an entry to the changelog please?

iamdustan:
Thanks for making me add the tests @basecode.

I had written them in backwards (the method names are in reverse order as to what I am used to).

src/tools.js
```diff
@@ -339,7 +339,34 @@ define([], function() {
...
+    /**
+     * Get degrees value from a radian.
+     *
+     * @param {Number} rad radians
+     * @param {Number} [precision] Floating point precision
+     * @returns {String} degrees
+     */
+    degFromRad: function (rad, precision) {
+      var degrees = rad * 180 / Math.PI;
+      if (precision) {
```
basecode: A precision of `0` would fail. Could you please add according tests?

You just want to check for "undefined" and "null". Use this instead: 

```
// checks for undefined and null
if (precision != null) {
```

src/tools.js
```diff
@@ -339,7 +339,34 @@ define([], function() {
...
+
+    /**
+     * Get degrees value from a radian.
+     *
+     * @param {Number} rad radians
+     * @param {Number} [precision] Floating point precision
+     * @returns {String} degrees
+     */
+    degFromRad: function (rad, precision) {
+      var degrees = rad * 180 / Math.PI;
```
basecode: Please cache the property lookup of `Math.PI` at the beginning of the page and use the cached variable instead.

```
// cache property lookup
var PI = Math.PI;
```

src/tools.js
iamdustan: I did a check for `typeof precision === 'number'` since any other type would through a SyntaxError.

src/tools.js
iamdustan: Fixed.

test/tools-spec.js
```diff
@@ -284,4 +284,53 @@ define([
...
+
+      expect(degFromRad(radians)).toBe(expected);
+    });
+
+    it('converts radians to ndegrees with a precision of 0', function () {
+      var radians = 0.785398163; // 45 degrees
+      var expected = toDegrees(radians).toFixed(0) + 'deg';
+      console.log('expected', expected);
+
+      expect(degFromRad(radians, 0)).toBe(expected);
```
basecode, line 331: +1
- Please remove `console.log`
- Would be good if you could be more explicit. Instead of  `toDegrees(radians).toFixed(0) + 'deg'` just write down your expectation, which is `45deg`. Hard to verify otherwise. Then you could also get rid of  `toDegrees()` and the comments.

iamdustan, line 331: Haha. Shoot! Who would’ve thought such a _simple_ addition could be done so sloppily.

iamdustan:
I like the `xToY` approach. The x is for verbosity and the y is what I want, being the last thing read it keeps it closer to mind.

To implode this pull request: in addition to the discussion about wording we had talked about css multiple units `deg` `rad` `turn` and `grad` (https://developer.mozilla.org/en-US/docs/CSS/angle). Would this pull request be usurped by that? The only real usefulness in `radToDeg` is to log out numbers in mere-mortal form.'

``` javascript
/**
 * @param {String} angle the angle in grad, deg, turn or rad
 * @param {String} returnUnit the unit to return
 */
function parseAngle(angle, returnUnit) {}
```

iamdustan:
Hmm... the motivation behind returnUnit is simply to get the angle in degrees back, but I don’t want that to be the motivating factor behind the increased complexity of that method so I’m open to ditching it.

basecode:
Shall we remove the returnUnit and precision-parameter and just return the computed value? I think that would make everything more clear. When it's done we can tackle the unit issue.

iamdustan:
What do you think about keeping these? I’m cool with ditching and closing this in favor of #167

basecode:
Closed in favor of #167

PR #167, opened 2012-12-05
tools.parseUnitString and tools.parseAngle

iamdustan:
Building upon these previous conversations, here are some converting methods to start accepting angles in any format throughout the project.

https://github.com/uxebu/bonsai/pull/142#issuecomment-9942288
https://github.com/uxebu/bonsai/pull/165#issuecomment-10999392

+@basecode

### DIFF
--- a/src/tools.js
+++ b/src/tools.js
@@ -349,7 +349,7 @@ define([], function() {
      * @returns {Object} { amount, unit }
      */
     parseUnitString: function (str) {
-      var amount = parseFloat(str.match(/^\-?\d+/)),
+      var amount = parseFloat(str),
           unit = str.match(/[a-z]*$/i)[0];
 
       return { amount: amount, unit: unit };
@@ -363,19 +363,19 @@ define([], function() {
      * @returns {Number} The angle in radians
      */
     parseAngle: function(angle) {
-      var parts = tools.parseUnitString(angle), unit = parts.amount, radians;
+      var parts = tools.parseUnitString(angle), amount = parts.amount, radians;
 
       switch (parts.unit) {
         case 'deg':
-          radians = unit * PI / 180; break;
+          radians = amount * PI / 180; break;
         case 'grad':
-          radians = unit * PI / 200; break;
+          radians = amount * PI / 200; break;
         case 'turn':
-          radians = unit * 2 * PI; break;
+          radians = amount * 2 * PI; break;
         case 'rad':
-          radians = unit; break;
+          radians = amount; break;
         default:
-          return 0;
+          radians = 0;
       }
 
       return radians;
--- a/src/tools.js
+++ b/src/tools.js
@@ -345,14 +345,18 @@ define([], function() {
     /**
      * Parse a given unit string into the amount and unit
      *
-     * @param {String} any The unit string. E.g. '90deg', '7em'
+     * @param {any} any Anything that can be converted to a string. E.g. '7deg', ['50%']
      * @returns {String} unit
      */
-    parseUnitString: function (any) {
+    extractUnit: function (any) {
+      if (typeof any === 'undefined' || typeof any === 'number') {
+        return '';
+      }
+
       // make sure it's a string and remove trailing whitespace
       var unit = String(any).replace(/\s+$/, '');
       // returns extracted unit or empty string
-      return unit.match(/[a-z]*$/i)[0];
+      return unit.match(/[a-z%]*$/i)[0];
     },
 
     /**
@@ -364,7 +368,7 @@ define([], function() {
      */
     parseAngle: function(angle) {
       var radians,
-          unit = tools.parseUnitString(angle),
+          unit = tools.extractUnit(angle),
           amount = parseFloat(angle);
 
       switch (unit) {
--- a/src/tools.js
+++ b/src/tools.js
@@ -345,35 +345,38 @@ define([], function() {
     /**
      * Parse a given unit string into the amount and unit
      *
-     * @param {String} str The unit string. E.g. '90deg', '7em'
-     * @returns {Object} { amount, unit }
+     * @param {String} any The unit string. E.g. '90deg', '7em'
+     * @returns {String} unit
      */
-    parseUnitString: function (str) {
-      var amount = parseFloat(str),
-          unit = str.match(/[a-z]*$/i)[0];
-
-      return { amount: amount, unit: unit };
+    parseUnitString: function (any) {
+      // make sure it's a string and remove trailing whitespace
+      var unit = String(any).replace(/\s+$/, '');
+      // returns extracted unit or empty string
+      return unit.match(/[a-z]*$/i)[0];
     },
 
     /**
      * Parses the angle string to radians
      * https://developer.mozilla.org/en-US/docs/CSS/angle
      *
-     * @param {String} angle The angle/unit string
+     * @param {Number|String} angle The angle/unit string
      * @returns {Number} The angle in radians
      */
     parseAngle: function(angle) {
-      var parts = tools.parseUnitString(angle), amount = parts.amount, radians;
+      var unit = tools.parseUnitString(angle),
+          amount = parseFloat(angle),
+          radians;
 
-      switch (parts.unit) {
+      switch (unit) {
+        case '': // default
+        case 'rad':
+          radians = amount; break;
         case 'deg':
           radians = amount * PI / 180; break;
         case 'grad':
           radians = amount * PI / 200; break;
         case 'turn':
           radians = amount * 2 * PI; break;
-        case 'rad':
-          radians = amount; break;
         default:
           radians = 0;
       }
--- a/src/tools.js
+++ b/src/tools.js
@@ -363,9 +363,9 @@ define([], function() {
      * @returns {Number} The angle in radians
      */
     parseAngle: function(angle) {
-      var unit = tools.parseUnitString(angle),
-          amount = parseFloat(angle),
-          radians;
+      var radians,
+          unit = tools.parseUnitString(angle),
+          amount = parseFloat(angle);
 
       switch (unit) {
         case '': // default
--- a/src/tools.js
+++ b/src/tools.js
@@ -349,10 +349,6 @@ define([], function() {
      * @returns {String} unit
      */
     extractUnit: function (any) {
-      if (typeof any === 'undefined' || typeof any === 'number') {
-        return '';
-      }
-
       // make sure it's a string and remove trailing whitespace
       var unit = String(any).replace(/\s+$/, '');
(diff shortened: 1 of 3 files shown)


src/tools.js
```diff
@@ -339,6 +340,45 @@ define([], function() {
...
+    },
+
+    /**
+     * Parse a given unit string into the amount and unit
+     *
+     * @param {String} str The unit string. E.g. '90deg', '7em'
+     * @returns {Object} { amount, unit }
+     */
+    parseUnitString: function (str) {
+      var amount = parseFloat(str.match(/^\-?\d+/)),
```
basecode: I think parseFloat works quite well without the regex.

src/tools.js
```diff
@@ -339,6 +340,45 @@ define([], function() {
...
+
+    /**
+     * Parses the angle string to radians
+     * https://developer.mozilla.org/en-US/docs/CSS/angle
+     *
+     * @param {String} angle The angle/unit string
+     * @returns {Number} The angle in radians
+     */
+    parseAngle: function(angle) {
+      var parts = tools.parseUnitString(angle), unit = parts.amount, radians;
```
basecode: Shouldn't it be `amount = parts.amount` instead of `unit`?

basecode:
I like the idea of supporting units. `tools.parseAngle` sounds good. Not sure if we need `tools.parseUnitString` though.

src/tools.js
iamdustan: haha. very good point. It was early when I did that, I swear.

src/tools.js
iamdustan: Yes, yes it should.

iamdustan:
`parseUnitString` was meant to be more low level. Currently it’s motivated to be the string-to-parts work of `parseAngle` and letting parseAngle really just deal with the conversion, but other unit types have different units of measure. E.g. `1s` `1000ms` `3min` `50%`

basecode:
That makes sense.

src/tools.js
```diff
@@ -339,6 +340,45 @@ define([], function() {
...
+          unit = str.match(/[a-z]*$/i)[0];
+
+      return { amount: amount, unit: unit };
+    },
+
+    /**
+     * Parses the angle string to radians
+     * https://developer.mozilla.org/en-US/docs/CSS/angle
+     *
+     * @param {String} angle The angle/unit string
```
basecode: `parseAngle` should also allow numbers (that are handled as radians). That would allow us to parse the parameter independent from the type. `var radians = tools.parseAngle(param);`

iamdustan: Good catch.

Simple enough as `if (typeof angle === 'number') return angle;` or should there be more robust handling of sorts (not sure what that would be...)

basecode: I'm tempted to let `parseUnitString` only return the unit and not "amount". For three reasons: 
1) `parseFloat` is nothing we need to put in a helper function. It is already a well tested helper function itself and can safely be used. 
2) We don't need to allocate another object (to be able to return 2 values)
3) It saves use from using if-else

Code:

```
extractUnit: function (any) {
  // make sure it's a string and remove trailing whitespace
  var unit = String(any).replace(/\s+$/, '');
  // returns extracted unit or empty string
  return unit.match(/[a-z]*$/i)[0];
}
```

```
parseAngle: function(angle) {

  var radians;
  var amount = parseFloat(angle);
  var unit = extractUnit(angle);

  switch(unit) {
    case 'deg': 
      radians = amount * PI / 180;
      break;
    …
    case '': // default
    case 'rad':
      radians = amount;
      break;
  }

  return radians;

}

```

What do you think?

iamdustan: :thumbsup:

I don’t know why I had completely blanked and not thought of 1. I also disliked it very much for number two so that is  a big one. Should we handle someone having white space between the number and unit as well?

`'7 deg'`?

basecode: `extractUnit("7 deg")` //  "deg"
That's fine, isn't it? Your regex does already the work of ignoring leading whitespace.

iamdustan:
:thumbsup: Any other cases we can think of to test and handle?

src/tools.js
```diff
@@ -339,6 +340,48 @@ define([], function() {
...
+
+    /**
+     * Parses the angle string to radians
+     * https://developer.mozilla.org/en-US/docs/CSS/angle
+     *
+     * @param {Number|String} angle The angle/unit string
+     * @returns {Number} The angle in radians
+     */
+    parseAngle: function(angle) {
+      var unit = tools.parseUnitString(angle),
```
basecode: Please put `radians` at the beginning. Variables that are not explicitly assigned should be listed first. I know there aren't any coding-styles available yet for bonsai.. but soon :)

iamdustan:
......annnnnnnnd done. I hadn’t even realized that nuance in the bonsai style. I’ve been trying to keep my additions like they belong in the family :)

src/tools.js
```diff
@@ -339,6 +340,48 @@ define([], function() {
...
       return array;
+    },
+
+    /**
+     * Parse a given unit string into the amount and unit
+     *
+     * @param {String} any The unit string. E.g. '90deg', '7em'
+     * @returns {String} unit
+     */
+    parseUnitString: function (any) {
```
basecode: Since `parseUnitString` can not just parse strings but also arrays (`parseUnitString(["7em"])`) for example wouldn't it be better to rename `parseUnitString` to `extractUnit` and change the JSDOC's "@ param" to `{any}`?

src/tools.js
```diff
@@ -339,6 +340,48 @@ define([], function() {
...
+        case '': // default
+        case 'rad':
+          radians = amount; break;
+        case 'deg':
+          radians = amount * PI / 180; break;
+        case 'grad':
+          radians = amount * PI / 200; break;
+        case 'turn':
+          radians = amount * 2 * PI; break;
+        default:
```
basecode: Why do we need "default:" here?

test/tools-spec.js
```diff
@@ -284,4 +284,58 @@ define([
...
+      expect(returned).toEqual(expected);
+    });
+
+    it('should work with trailing whitespace', function() {
+      var returned = parseUnitString('-741232grad    ');
+      var expected = 'grad';
+
+      expect(returned).toEqual(expected);
+    });
+
```
basecode, line 312: Other edge cases: 
- should work with leading whitespace
- should return an empty string in case of a number

test/tools-spec.js
```diff
@@ -284,4 +284,58 @@ define([
...
+      expect(calculated).toBeCloseTo(expected, 10);
+    });
+
+    it('calculates radians for gradians', function () {
+      var calculated = parseAngle('-32grad');
+      var expected = -0.5026548245743669;
+
+      expect(calculated).toBeCloseTo(expected, 10);
+    });
+
```
basecode, line 365: Other edge cases:
-  calculates radians for radians as a string
-  calculates radians for radians as a number

src/tools.js
iamdustan: Yup. That does make sense. I didn’t even think of the example of passing an array in. Update coming.

src/tools.js
iamdustan: Don’t all `switch` statements need a default? I don’t often use switch statements. I usually do the dirty super ternary instead :|

``` javascript
var radians = 'rad' === unit ? amount
            : 'deg' === unit ? amount * PI / 180
            : 'grad' === unit ? amount * PI / 200
            : 'turn' === unit ? amount * 2 * PI
            : 0;
```

test/tools-spec.js
iamdustan, line 365: for radians as a string as in `parseAngle('0.321512324')` or as in `parseAngle('0.321512341rad')`?

iamdustan, line 365: Added tests for the following scenarios:

`parseAngle('0.42131')` `parseAngle(0.42131)` and `parseAngle('0.42131rad')`

test/tools-spec.js
iamdustan, line 312: Tests for both of those as well as the unit being `%` and for when undefined is passed in.

src/tools.js
```diff
@@ -339,6 +340,52 @@ define([], function() {
...
+    },
+
+    /**
+     * Parse a given unit string into the amount and unit
+     *
+     * @param {any} any Anything that can be converted to a string. E.g. '7deg', ['50%']
+     * @returns {String} unit
+     */
+    extractUnit: function (any) {
+      if (typeof any === 'undefined' || typeof any === 'number') {
```
basecode: I don't think that we need that check. What about `"", null, [], {}, …`? I see your point but I don't think that we should expect `undefined` and `number` of all types. The cast in L357 takes care of invalid values.

iamdustan: so do some test after `var unit = ...` to make sure we return the empty string if given a poor argument?

```
String(undefined) //=> "undefined"
String(null) //=> "null"
String({}) //=> "[object Object]"
String([]) //=> ''
String(['singleDepth']) //=> 'singleDepth'
String(['bigger', 'array', 4]) //=> 'bigger,array,4'
```

src/tools.js
basecode: Well, I think a switch statement don't need necessarily a "default", especially when the default value was previously set (in L470). You already implicitly assigned undefined.

src/tools.js
basecode: No. That's sth the author/dev need to fix. That's why I suggested to remove the type-checks.

basecode:
Oh and please add a line to the changelog :)

basecode:
And thanks for your endurance! This is going to be a really nice feature!

test/tools-spec.js
basecode, line 365: +1

src/tools.js
iamdustan: What do you think of reversing it? Rather than `case ''` falling through to `case 'rad'` we let `'rad'` fail through to default.

src/tools.js
iamdustan: :thumbsup: I can roll with that. Removed the checks, tests are still passing.

iamdustan:
The endurance is easy over here! I built it up working on matrices, getters/setters, and compounding transformations that I only half understand.

iamdustan:
Pushed.

src/tools.js
basecode: In that case we wouldn't even need `case 'rad'`.  But `parseAngle('2whatever')` would return `2 (rad)`.  :(

basecode:
Thanks for this addition @iamdustan ! :+1:

iamdustan:
:) We made it!

klipstein:
Awesome!

PR #168, opened 2012-12-07
Prevent dom clutter

pvdz:
Removes `transform` for the identity matrix and `pointer-events` for the default of `inherit`. Maybe more props can be prevented but I think some are used elsewhere.

### DIFF
--- a/src/renderer/svg/svg.js
+++ b/src/renderer/svg/svg.js
@@ -57,19 +57,20 @@ define([
       fontPrefix = AssetController.handlers.Font.prefix;
 
   var basicAttributeMap = {
-    cap: 'stroke-linecap',
-    join: 'stroke-linejoin',
-    miterLimit: 'stroke-miterlimit',
-    opacity: 'opacity',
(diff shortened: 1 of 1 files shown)


src/renderer/svg/svg.js
```diff
@@ -20,6 +20,22 @@ define([
...
 
+  /**
+   * Sets a style property on a style object. Avoids unnecessary creation of
+   * style attributes in the DOM, to ease debugging.
+   *
+   * @param {CSSStyleDeclaration} style The style object to set the property on
+   * @param {string} name The name of the property to set
+   * @param {string} value The value of the property to set.
+   */
+  function setStyle(style, name, value) {
```
basecode, line 31: Is closured. How are you going to test that function?

davidaurelio, line 31: I’m not. It uses 3rd-Party API and is part of the implementation, not the public interface

basecode, line 31: Yep, you're right!

PR #171, opened 2012-12-12
Add `event.relatedTarget` for mouseover and mouseout events

davidaurelio:
This adds a `relatedTarget` property to mouseover and mouseout event objects. It refers to the bonsai element that has been left or accessed.

### DIFF
--- a/example/library/movies/movie_list.js
+++ b/example/library/movies/movie_list.js
@@ -73,6 +73,7 @@ movieList = {
     'event-multi-touch.js',
     'event-multi-touch2.js',
     'event-mouse-button.js',
+    'event-mouse-over-out.js',
     'key-events.js'
   ],
   'Filter': [
--- a/src/renderer/svg/svg_event_handlers.js
+++ b/src/renderer/svg/svg_event_handlers.js
@@ -47,7 +47,7 @@ define([
    * @return {Node}
    */
   function findBonsaiObject(node) {
-    while (node && (node.nodeType !== ELEMENT_NODE || node.hasAttribute('bs-data-id'))) {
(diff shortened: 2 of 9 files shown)


example/library/movies/movie_list.js
```diff
@@ -73,6 +73,7 @@ movieList = {
```
klipstein, line 72: why do we have to maintain two movie lists?

src/renderer/svg/svg_event_handlers.js
```diff
@@ -16,6 +19,40 @@ define([
...
   }
 
+  /**
+   * Returns the bonsai id of a DOM node
+   *
+   * @param {Node} node A DOM node
+   * @return {number} The bonsai id of the dom node or -1
+   */
+  function getBonsaiIdOf(node) {
+    var id = node && node.nodeType === ELEMENT_NODE && node.getAttribute('data-bs-id');
```
klipstein, line 29: On https://github.com/uxebu/bonsai/pull/171/files#L5R50 you use `bs-data-id` and here `data-bs-id`. Is that correct?

example/library/movies/movie_list.js
davidaurelio, line 72: let me find out

src/renderer/svg/svg_event_handlers.js
davidaurelio, line 29: of course not. Good catch. Thank you!

example/library/movies/movie_list.js
davidaurelio, line 72: example/profiling uses movie_list.js, example/library the json file :(

72790428ae787d3fc8074b33358ab004c3f775d1 did not change all of them.

PR #172, opened 2012-12-12
Tree traversal

davidaurelio:
- the emission of “tick” / “advance” / frame number events now happens in tree order (depth first) rather than insertion order.
- it also adds a “tickEnd” event for each timeline after all child timelines emitted their frame events.
- There is still optimization opportunity: `addChildren()` could maintain a child timeline array on each timeline instance to make iteration faster.

### DIFF
--- a/src/runner/display_list.js
+++ b/src/runner/display_list.js
@@ -150,22 +150,6 @@ define([
     }
   };
 
-  function activate(stage) {
-    DisplayObject.prototype._activate.call(this, stage);
-    var children = this.displayList.children;
-    for (var i = 0, length = children.length; i < length; i += 1) {
-      children[i]._activate(stage);
-    }
-  }
-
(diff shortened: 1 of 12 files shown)


iamdustan:
This is definitely lower level than I’ve traveled and so I don’t feel qualified to judge it. To keep the pull request from sitting for another 6 months is there anyone else on the @uxebu team willing? With @wolframkriesing’s recent twitter stream I can definitely see you guys have been busy with pixelplant.

wolframkriesing:
thanks for triggering this again :)
sorry if we are slow.

## 

Kind regards / Saludos / Mit freundlichen Grüßen

Wolfram Kriesing - CTO, Co-Founder

Flash on the iPad? We make it work!
kriesing@uxebu.com, mobile: +49 174 300 4595
uxebu Inc.

PR #176, opened 2012-12-18
Revert element id string resolution to happen at `run`

iamdustan:
See comments here: https://github.com/uxebu/bonsai/pull/152#issuecomment-11142897


davidaurelio:
Will merge this now, and try to think of a nicer way for the test. Thank you.

Issue #84, opened 2012-09-26
Text positioning api


pvdz:
``` js
new Text('foo').attr({x,y,width:500,height:300,valign:'top',halign:'middle'})
```

There would be two new attributes: `halign` and `valign` which control horizontal and vertical alignment respectfully. Valid values for horizontal would be `'left','middle','right'`, and vertical `'top','middle','bottom'` (`middle` comes from css, i think it'd be nice if `center` would also map to `middle` here).

I know text has way more vertical lines to align to, but these three are the most important bits.

The alignment should (probably?) be relative to the bounding box of the `Text` element. So if you specify a `Text` with `width=500` and `halign='right'` then the _right-hand side_ of the text should be at 500px from the left-hand-side of the text element. Might be obvious but just pointing it out anyways :)

basecode:
Good point. Related: https://github.com/uxebu/bonsai/tree/text-align

there4:
+1

I've also noticed that the text api also seems to always return zeroes for the getBoundingBox. It appears that this is not implemented, and is just returning the parent display object method.

basecode:
@there4, you're right. The reason for that is that it isn't trivial. I would suggest to let the author (manually) define a TextBox (x, y, width, height) instead.

It's not trivial because:
- There is no API (e.g. DOM) available in the runner context (where all bonsai movies are executed) that could do the job for us
- We don't have the numbers. A glyph's outline depends on the inner-bbox, hinting, kerning, ligatures… We don't have access to these numbers in case of system fonts.
- Even if we would have access to the font's source-file, it would probably take some time to get the numbers correct.

Feel free contributing!

iammerrick:
I'm sorry the documentation for this piece is a bit sparse I am struggling to figure out how to simply accomplish this... :-)
![Screen Shot 2013-04-22 at 11 22 01 AM](https://f.cloud.github.com/assets/109635/410545/898d5f60-ab71-11e2-82f6-468b3f80c900.png)

basecode:
> I'm sorry the documentation for this piece is a bit sparse I am struggling to figure out how to simply accomplish this... :-)

It's sparse because textAlign is not supported so far. I created a pull-request, see #208.

washu:
could you not after the svg render draw the text/textspan, grab the bounding box to return the width,height back to the runner to atleast populate the getBoundingBox call?

basecode:
[...]

PR #188, opened 2013-01-23
Text#attr[textOrigin]

### LINKED ISSUE #166: Text positioned incorrectly in IE
The docs say that Text is anchored on the top left (meaning that the top left of the text object will be positioned at its origin), but in Firefox it's the bottom left.  For instance:

```
new Rect(0, 0, 50, 50).fill("red").addTo(stage);
var t = new Text('hi').attr('fontSize', '50px').attr('x', 0).attr('y', 0).addTo(stage);
```

In Chrome, the text is inside the rectangle.  In Firefox, it's above it.

Workaround: adding a css rule for "tspan { alignment-baseline: alphabetic; }" makes Chrome act the same as Firefox (incorrectly, but at least it gets them doing the same thing).  Firefox does not seem to respect the alignment-baseline attribute, which may be the cause of the problem.

basecode:
This pull request fixes #166 and "silently" adds support for three different textOrigins. Namely "top", "center" and "bottom".

### DIFF
--- a/src/renderer/svg/svg.js
+++ b/src/renderer/svg/svg.js
@@ -656,7 +656,7 @@ define([
     text.setAttributeNS(xlink, 'text-anchor', 'start');
 
     if (textOrigin != null) {
-      text.setAttribute('alignment-baseline', textOrigin === 'top' ? 'hanging' : '');
+      setStyle(style, 'alignmentBaseline', textOrigin === 'top' ? 'hanging' : '');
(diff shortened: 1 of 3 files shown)


basecode:
The reason why I propose "textOrigin" being a "silent" or non-public addition is because we're probably changing the behaviour of Text in the future due to #84. IMO making "textOrigin" public would conflict with the idea of exposing a "verticalAlign" API. Could be that the Text API ends up being kind of a TextBox API.

davidaurelio:
Thanks

PR #193, opened 2013-02-04
Add `DisplayObject#globalToLocal` and `DisplayObject#localToGlobal`

davidaurelio:


test/display_object-spec.js
```diff
@@ -409,5 +410,69 @@ define([
...
+    });
+
+    describe('localToGlobal', function() {
+      it('should apply the inversed transform of an display object to a point', function() {
+        var point = new Point(-102, 23.75);
+        var matrix = new Matrix(1.0625, 0.3125, -1.25, -1.09375, 26.5, -34);
+        var displayObject = new DisplayObject().attr('matrix', matrix);
+
+        expect(displayObject.localToGlobal(point))
+          .toEqual(matrix.invert().transformPoint(point));
```
basecode, line 453: It would make the test more readable, if you would write down the expected value explicitly.

davidaurelio, line 453: I will add some tests that test “target values” and make them very explicit. 

But I think for the existing tests here it makes more sense to get the idea how to get to the desired values, rather than just writing downs six floating point values. It would not be clear how they relate to the input values.

basecode, line 453: I don't really agree. We don't care about implementation details in tests. But we care about the expected value and I'm not able to "see" the value right now. I need to copy the test in orbit to see the result.

iamdustan, line 453: I think there is value in both approaches. I oftentimes read through tests to see how a project works and how to use it in code. Docs are great for API details and arguments, but seeing how things can actually be used together is much more valuable.

What if the test had something along the lines of:

```
var localToGlobalPoint = new Point(x, y); // displayObject.localToGlobal(point)
var expected = { x: expectedX, y: expectedY } // matrix.invert().transformPoint(point);
```

PR #194, opened 2013-02-12
Implemented rudimentary charCode to be returned by keydown event.

wolframkriesing:
Added a basic implementation of charCode to return for keydown events, as long as they are broken in DOM.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,8 @@
+v0.4.4
+-------------------
+* Added a basic implementation of charCode to return for
+  keydown events, as long as they are broken in DOM
+
 v0.4.3
 -------------------
 * Add support for `strokeDash` and `strokeDashOffset` attributes on all Paths.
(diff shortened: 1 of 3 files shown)


basecode:
Looks fine. Two questions:
- Does it make sense to expose both values? charcode and keycode
- What about `which`? http://stackoverflow.com/a/1445228

wolframkriesing:
i dont care if it is called `which` or `keyCode` in the end, i think `keyCode` is more explicit

basecode:
The wording wasn't subject of my question. I asked why we expose `charcode` and `keycode` and why we ignore `event.which` (see link).

wolframkriesing:
tbh the article confuses me, which is the same as keycode in some cases, in that case i think `charCode` and `keyCode` are the best choices, they clearly state that they are different and what the mean (maybe I am a bit flash biased). but i see no value in `which`, and esp. that it could mean a lot and nothing

davidaurelio:
Closed in favor of #201

PR #195, opened 2013-02-22
Pointer events cleanup and alignment

davidaurelio:
- Cleans up event handling in the svg renderer.
- adds “deltaX/Y” and “diffX/Y” to regular “pointermove” events as well (not only drags) when triggered by mouse events.
- adds “deltaX/Y” to “pointermove” and “drag” events when triggered by touches
- does not send unnecessary drag events to the runner.
- Stops memory leaking from touch events.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -2,6 +2,7 @@ v0.4.3
 -------------------
 * Add support for `strokeDash` and `strokeDashOffset` attributes on all Paths.
 * Add `DisplayObject#globalToLocal()` and `DisplayObject#localToGlobal()`
+* Consistent delivery of diffX/Y and deltaX/Y offsets for move and drag events
 
 v0.4.2 / 2013-01-14
 -------------------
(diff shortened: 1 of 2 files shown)


basecode:
Thanks, looks good. What about tests?

iamdustan:
I can’t tell by glancing through this, but does it or would it make sense to include the diffX/diffY from the start coordinate for drag events rather than (or in addition to) the last event emission?

That way things like https://github.com/uxebu/bonsai-demos/blob/gh-pages/demos/draggable-shapes/index.html#L49 can be handled by bonsai and userland doesn’t need to keep track of that state?

davidaurelio:
Yes, absolutely. I’m already working on it in the next branch.
- diffX/Y will be offsets from the start of the drag
- deltaX/Y will be offsets relative to the last event (also fired for mousemove)

iamdustan:
Awesome. :thumbsup:

davidaurelio:
Closed in favor of #201

PR #201, opened 2013-03-06
Elements under pointer

davidaurelio:
- Contains #194 and #195 
  - Added a basic implementation of charCode to return for keypress events
  - Cleans up event handling in the svg renderer.
  - adds “deltaX/Y” and “diffX/Y” to regular “pointermove” events as well (not only drags) when triggered by mouse events.
  - adds “deltaX/Y” to “pointermove” and “drag” events when triggered by touches
  - does not send unnecessary drag events to the runner.
  - Stops memory leaking from touch events.
- breaks out dom event recognition and translation to bonsai events into own module
- adds experimental event feature: get all elements under the pointer for pointer events


wolframkriesing:
i can see that the test cases i added pass too, so for that part i can say +1, the rest i dont feel knowledgable enough.

PR #208, opened 2013-04-23
textAlign

basecode:
This pull-request introduces support for `textAlign` attribute on Text. It can be set via `attr` and allows one of the following values: "left", "center", "right". Default is "left". Partially addresses #84.

```
new Text('my Text').attr('textAlign', 'left');
```

**Why not "start" and "end"?**

It makes sense to implement "start" and "end" as soon as Bonsai also supports text direction (e.g. from right to left) similar to the [SVG](http://www.w3.org/TR/SVG/text.html#TextAnchorProperty) or [Canvas2d](http://www.whatwg.org/specs/web-apps/current-work/multipage/the-canvas-element.html#dom-context-2d-textalign) spec for example.


iamdustan:
Code looks clean as butter to me.
I just tested in latest Chrome, Safari and Firefox with consistent results.

:thumbsup: from me.

basecode:
Thanks for reviewing @iamdustan!

src/runner/text.js
```diff
@@ -78,6 +78,15 @@ define([
     return text.join('');
   }
 
+  function getTextAlign() {
+    return this._textAlign;
+  }
+  function setTextAlign(textAlign) {
+    if (/^(left|center|right)$/.test(textAlign)) {
```
klipstein: Why don't you use `indexOf`. It should be faster:

``` js
if (tools.indexOf(['left', 'center', 'right'], textAlign) > -1) {
  // ...
}
```

klipstein:
@basecode please also change the `Text`-documentation here https://github.com/uxebu/bonsai-docs/blob/gh-pages/overview/Text.md.

basecode:
Good point. I updated the Text documentation https://github.com/uxebu/bonsai-docs/pull/37.

src/runner/text.js
basecode: Although I don't consider this a "hotspot" I agree that the current approach could be optimised. I decided to go with a "primitive comparison". See: http://jsperf.com/regex-test-vs-array-indexof/2

klipstein: Definitely no hotspot, but thanks for changing it.

